PR #14, opened 2015-02-17
Totpauthenticator

wattale:


modules/jaggery-apps/portal/gadgets/user_profile/controllers/my-profile/edit-finish.jag
```diff
@@ -52,6 +52,9 @@ var userName = request.getParameter("user");
 
         for each (var f in fieldValues){
             var paramName = f.*::['claimUri'].text().toString();
+	    if(paramName=="http://wso2.org/claims/identity/encoding"||paramName=="http://wso2.org/claims/identity/qrcodeurl"){
```
isharak: Better to define these parameter names as constants and use here

JKAUSHALYA:
We are using TOTP connector as our authenticator now. Hence closing the PR.

PR #66, opened 2015-04-29
Add fido authenticator support for dashboard

isharak:


isharak:
Add fido dashboard support

PR #145, opened 2015-07-13
commented get decision deny test case

dracusds123:
please uncomment this once the issue is identified; user profile updated, but not reflected in the admin service call


johannnallathamby:
Closing PR since tests are not failing now.

PR #160, opened 2015-07-30
update dependency versions

hpmtissera:


hpmtissera:
Merge after all the repos are released (wss4j, balana, charon, rampart and amber)

PR #180, opened 2015-08-17
Config IS to start with different p2 profile. ex: workflow

harsha1979:


modules/distribution/src/assembly/bin.xml
```diff
@@ -259,6 +259,12 @@
             </includes>
         </fileSet>
 
+	<fileSet>
+            <directory>../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/dbscripts/bps</directory>
+            <outputDirectory>${pom.artifactId}-${pom.version}/dbscripts/bps</outputDirectory>
```
johannnallathamby: Check if this can be renamed to dbscripts/business-process to refer the feature repo instead of the product repo. If not lets suggest BPS to change it.

modules/distribution/src/assembly/bin.xml
```diff
@@ -540,6 +546,44 @@
             </source>
             <outputDirectory>wso2is-${pom.version}/repository/conf/identity</outputDirectory>
         </file>
+	<!--file>
+		    <source>../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/repository/conf/bps.xml
+		    </source>
+		    <outputDirectory>wso2is-${pom.version}/repository/conf/</outputDirectory>
+        	<fileMode>644</fileMode>
+        </file-->
```
johannnallathamby: Remove commented lines.

modules/distribution/src/assembly/bin.xml
```diff
@@ -540,6 +546,44 @@
...
+		    <outputDirectory>wso2is-${pom.version}/repository/conf/</outputDirectory>
+        	<fileMode>644</fileMode>
+        </file>
+
+       <file>
+            <source>
+            	src/repository/resources/bps/bps.xml
+	    </source>
+            <outputDirectory>wso2is-${pom.version}/repository/conf/</outputDirectory>
+        </file>
```
johannnallathamby: Can we have the bps.xml in distribution/conf folder? This is the usual place we have duplicated config files if necessary.

modules/distribution/src/repository/resources/bps/bps.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ ~ Copyright (c) 2005-2010, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ ~
```
johannnallathamby: Wrong license. Although it temporary fix it.

modules/p2-profile-gen/pom.xml
```diff
@@ -527,18 +541,428 @@
...
+
+
+					<execution>
+						<id>p2-profile-generation-for-workflow</id>
+						<phase>package</phase>
+						<goals>
+							<goal>p2-profile-gen</goal>
+						</goals>
+						<configuration>
+							<profile>workflow</profile>
```
johannnallathamby: Looks like workflow profile starts with all the bundles. How do we start with only BPS bundles, like a BPS backend server ? Can we add another profile ? I would suggest lets have 3 profiles.
1. default - carbon-identity bundles only
2. workflow - business-process bundles only
3. full - all bundles

johannnallathamby:
Have we installed the business-process features one by one and checked if we have got the minimum required features for IS to work with a embedded backend BPS server. We may not need all the backend features. We could be able to remove some of the backend feature which IS won't use.

Can you list the features that were installed in order to get the embedded BPS server running ?

pom.xml
```diff
@@ -1102,6 +1102,8 @@
         <cipher-tool.version>1.0.0-wso2v3</cipher-tool.version>
         <jacoco.agent.version>0.7.4.201502262128</jacoco.agent.version>
         <project.scm.id>my-scm-server</project.scm.id>
+        <carbon.business-process.version>4.4.2</carbon.business-process.version>
+	<carbon.analytics-common.version>1.0.0</carbon.analytics-common.version>
```
johannnallathamby: Fix indentation

PR #194, opened 2015-09-02
migration client for 5.0.0 to 5.1.0

madurangasiriwardena:


modules/distribution/migration/migration-5.0.0_to_5.1.0/wso2-is-migration-client/pom.xml
```diff
@@ -0,0 +1,218 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
johannnallathamby: Add license headers

PR #203, opened 2015-09-04
Fix for IDENTITY-3521

harsha1979:


modules/p2-profile-gen/pom.xml
```diff
@@ -780,19 +768,17 @@
                             <targetPath>
                                 file:${basedir}/target/wso2carbon-core-${carbon.kernel.version}/repository/components
                             </targetPath>
-                            <profile>workflow</profile>
+                            <profile>all</profile>
```
cdwijayarathna: Is this change older profile name 'workflow' to 'all', If so I think we should go for a more meaningful name than 'all'

PR #217, opened 2015-09-12
rename profile all to full

harsha1979:


cdwijayarathna:
I don't think 'full' makes more sense than 'all', both sounds same and also doesn't give any idea about the difference between default profile vs this profile,
also this will cause test failures, in tests also we have to change the profile name at WorkflowConstants.

PR #285, opened 2015-10-22
adding tenant domain field to playground sample

hpmtissera:


johannnallathamby:
Since we provide the authorization endpoint of IS as a input parameter we can  provide the tenant domain as a query parameter as well. Why do we need a separate input here ?

JKAUSHALYA:
Closing the PR until conflicts are resolved.

PR #334, opened 2015-11-19
Fixing - IDENTITY-4064

JKAUSHALYA:
https://wso2.org/jira/browse/IDENTITY-4064

### DIFF
--- a/modules/jaggery-apps/portal/gadgets/user_auth_apps/js/gadget.js
+++ b/modules/jaggery-apps/portal/gadgets/user_auth_apps/js/gadget.js
@@ -15,10 +15,18 @@ function drawPage() {
     if (json != null) {
 
         if (isArray(json.return)) {
+
             for (var i in json.return) {
+
+                var username = json.return[i].username;
+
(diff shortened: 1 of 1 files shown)


modules/jaggery-apps/portal/gadgets/user_auth_apps/js/gadget.js
```diff
@@ -15,10 +15,18 @@ function drawPage() {
     if (json != null) {
 
         if (isArray(json.return)) {
+
             for (var i in json.return) {
+
+                var username = json.return[i].username;
+
+                if (username.indexOf("carbon.super") > -1) {
```
Pushpalanka: carbon.super should be available as a constant. Can re-use it. So how would this work in tenant mode? There are util methods available to getTenantAwareUserName to seperate user name from tenant domain.

JKAUSHALYA: Since this is a Javascript, constants are not available. We only need to remove the tenant domain name if it is only the super tenant, hence checking for the "carbon.super".

PR #364, opened 2015-12-07
Fixing - IDENTITY-4198

JKAUSHALYA:
https://wso2.org/jira/browse/IDENTITY-4198

### DIFF
--- a/modules/jaggery-apps/portal/gadgets/connected_accounts/controllers/service-clients/identityProviderMgtClient.jag
+++ b/modules/jaggery-apps/portal/gadgets/connected_accounts/controllers/service-clients/identityProviderMgtClient.jag
@@ -3,20 +3,21 @@
 var log = new Log();
 
 var util = require("../../util/utility.jag");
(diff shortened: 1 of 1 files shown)


modules/jaggery-apps/portal/gadgets/connected_accounts/controllers/service-clients/identityProviderMgtClient.jag
```diff
@@ -3,20 +3,21 @@
 var log = new Log();
```
amilam: When defining log object in jagger, its better to define it with the file path. Then, we know which jaggery file printed that log. For example,

var log = new Log("site/pages/index.jag").

Make this a practice for all the jaggery codes.

JKAUSHALYA: Fixed in https://github.com/wso2/product-is/pull/410

PR #408, opened 2016-01-20
Fixing CODE_ID not null constraint

madurangasiriwardena:


modules/migration/migration-5.0.0_to_5.1.0/wso2-is-migration-client/src/main/java/org/wso2/carbon/is/migration/client/MigrateFrom5to510.java
```diff
@@ -452,31 +453,50 @@ public void migrateIdentityDBFinalize(){
...
 
-            notNullPS = identityConnection.prepareStatement(alterTokenIdNotNull);
-            notNullPS.execute();
+            tokenIdNotNullPS = identityConnection.prepareStatement(alterTokenIdNotNull);
+            tokenIdNotNullPS.execute();
 
             primaryKeyPS = identityConnection.prepareStatement(setAccessTokenPrimaryKey);
             primaryKeyPS.execute();
 
+            codeIdNotNullPS = identityConnection.prepareStatement(alterCodeIdNotNull);
```
vidulkeyt: In our test DB there are still 6k records with code_id = null (we have migrated it w/o this constraint)
So I assume the fix will fail again when perform this alter
Probably you should delete/update data in IDN_OAUTH2_AUTHORIZATION_CODE table before switching the constraint on.

madurangasiriwardena: We already add an ID to the CODE_ID column from the migration

PR #409, opened 2016-01-28
Added the test case for IDENTITY-4291

JKAUSHALYA:
**Don't merge this until the Carbon next version released.**


JKAUSHALYA:
Closing since we can have a better test solution for this use case without uploading an existing database.

PR #444, opened 2016-03-16
Test case user store count feature

Pushpalanka:


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/user/store/JDBCUserStoreAddingTestCase.java
```diff
@@ -200,8 +208,37 @@ public void changePassWordByUserTest() throws Exception {
 
     }
 
-
     @Test(groups = "wso2.is", dependsOnMethods = "addUserIntoJDBCUserStore")
```
cdwijayarathna: Why these methods need to execute sequentially?

cdwijayarathna:
can we squash these commits into 1 commit with name "test case user store count feature"

PR #454, opened 2016-04-04
Passive sts SAML2.0 feature support sample and autotests

cgchamath:
Updated the PassiveSTSSample to send SAML2.0 token requests.

Testing the feature
--------------------------

Deploy the PassiveSTSSample application.
To request a SAML 2.0 token use the URL - http://localhost:8080/PassiveSTSSampleApp?samlv=2-0
This will generate a passive sts request by populating the wreq parameter to have the requested token type to be 2.0. 
In the same way you can request SAML 1.1 token with the URL - http://localhost:8080/PassiveSTSSampleApp/?samlv=1-1
If you did not specify any parameter by default it will generate a SAML 1.1 token.


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/sts/TestPassiveSTS.java
```diff
@@ -1,9 +1,18 @@
 package org.wso2.identity.integration.test.sts;
 
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
```
Pushpalanka: According to code styling java package imports should be below.

modules/samples/passive-sts/passive-sts-client/PassiveSTSFilter/src/main/java/org/wso2/identity/passivests/sample/AuthFilter.java
```diff
@@ -3,6 +3,7 @@
 import static org.apache.commons.lang.StringEscapeUtils.escapeHtml;
 
 import java.io.IOException;
+import java.net.URLEncoder;
```
Pushpalanka: java packages should come at lowest

cgchamath: Fixed

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/sts/TestPassiveSTS.java
cgchamath: Fixed

PR #550, opened 2016-07-03
IDENTITY-3591 : Required changes for supporting challenge questions internalization.

mefarazath:
Changes required for https://wso2.org/jira/browse/IDENTITY-3591

Merge after : https://github.com/wso2-extensions/identity-governance/pull/32


modules/jaggery-apps/portal/gadgets/account-recovery/index.jag
```diff
@@ -68,11 +68,11 @@ session.put('user', user);
             }
         }
 
-print( orderedChallanges );
-    }catch(e){
-print(e);
+        print(orderedChallanges);
+    } catch (e) {
+        print(e);
```
darshanasbg: Do we need to print e directly? Shouldn't it print some general error to show in the UI and log the actual error?

PR #609, opened 2016-07-18
Adding integration test for IDENTITY-4776 Fix

grainier:
This test recreates the scenario using;
1. Add a secondary user store (with SCIMEnabled).
2. Add a user/role to above secondary user store.
3. Register an OAuth application, and generate a token (with primary store user).
4. Login to carbon using secondary store user (to invoke BasicAuthHandler).
5. Try to create a new user using SCIM Service (using auth token generated in step 3).

Please review and merge.


mefarazath:

PR #611, opened 2016-07-18
IDENTITY-4796 - Add test function to retrive federated IdPs for users who have login …

isurusuranga:
…permission


mefarazath:

PR #648, opened 2016-08-05
Front End improvements for IDENTITY-4832

IndunilRathnayake:


darshanasbg:
It was decided to avoid changing the UI, rather revoke consent from the revoke application itself.

PR #668, opened 2016-08-18
MongoDB User Store feature added 

asanthamax:
modified parent pom.xml and p2-profile-gen pom.xml file to to add MongoDB User Store


JKAUSHALYA:
Closing until conflicts are resolved.

PR #684, opened 2016-08-28
updated dependency version and test case to fix https://wso2.org/jira/browse/IDENTITY-4844

jenananthan:


mefarazath:

PR #770, opened 2016-10-21
Added "Account Association" services

ruwanta:


johannnallathamby:
Merged with #777 due to folder structure changes

PR #820, opened 2016-11-23
Clean up startup logs

mefarazath:
Change org.spring.framework log level to WARN in order to get rid of INFO logs that are less useful printed in the startup log.


darshanasbg:
Thanks @mefarazath for the contribution.

PR #835, opened 2016-12-02
Fix test cases

madurangasiriwardena:
If the resident idp is updated without properly setting idp properties, it will override the expected properties in the resident idp. Without these properties, authentication flow will not work properly.

There should be a array size check for the below code, as there can be ArrayIndexOutOfBoundsExceptions.
https://github.com/wso2-extensions/identity-governance/blob/v0.1.15/components/org.wso2.carbon.identity.recovery/src/main/java/org/wso2/carbon/identity/recovery/util/Utils.java#L286


darshanasbg:
Thanks @madurangasiriwardena for the contribution. Much appreciated..!

PR #871, opened 2016-12-13
Add migration for default governance connector property values

mefarazath:
$subject is required for the recovery flows to work after migrating from 5.2.0 to 5.3.0


modules/migration/migration-5.2.0_to_5.3.0/wso2-is-migration-client/src/main/java/org/wso2/carbon/is/migration/dao/IdpMetaDataDAO.java
```diff
@@ -0,0 +1,190 @@
...
+    public static IdpMetaDataDAO getInstance() {
+
+        return idpMetaDataDAO;
+    }
+
+
+    public int getResidentIdpId(int tenantId) throws ISMigrationException {
+
+        // we use the "RememberMeTimeout" attribute as a way to identify the resident IDP id.
+        String sql = "SELECT IDP_ID FROM IDP_METADATA WHERE TENANT_ID=? AND NAME=\"RememberMeTimeout\";";
```
pulasthi7: AFAIK, IDP.Name is 'LOCAL' for resident IdPs. We can use that instead of this property.

mefarazath: +1

PR #1014, opened 2017-01-17
IDENTITY-5592 change the extension to yaml

denuwanthi:


kasunbg:
Do we not have to modify the code that reads this conf file?

PR #1022, opened 2017-01-18
Code Improvements for self sign-up in user portal

IndunilRathnayake:


IndunilRathnayake:
@manuranga Thanks for your comments. Fixed all the issues with https://github.com/wso2/product-is/pull/1053

PR #1057, opened 2017-01-25
IDENTITY-5574: Adding form-post support to playground2 app

GDRDABARERA:


modules/samples/oauth2/playground2/src/main/webapp/oauth2-authorize-user.jsp
```diff
@@ -33,6 +34,10 @@
         if(usePKCE) {
             session.setAttribute(OAuth2Constants.OAUTH2_USE_PKCE, usePKCE);
         }
+        boolean formPostMode = formPostParameter != null && "yes".equals(formPostParameter);
```
pulasthi7: null check is redundant

PR #1126, opened 2017-02-14
IDENTITY-5725 fix Login Page error message should  be changed when us…

denuwanthi:
…er didin't enter a user name or a password.


portal/apps/org.wso2.carbon.iam.user-portal.app/src/main/pages/login.js
```diff
@@ -43,6 +43,9 @@ function getPrimaryDomainName(env) {
 
 function authenticate(username, password, domain) {
     try {
+        if(!username || !password){
```
GayanM: if(!(username && password))

PR #1207, opened 2017-03-31
Fixing internationalization issues

GayanM:


portal/apps/org.wso2.carbon.iam.user-portal.app/src/main/lang/en_US.properties
```diff
@@ -171,4 +171,4 @@ recover.username=Recover Username
 fill.below.to.recover=Please fill in the below form to recover your username
 password.history.validation.error.1000=Password you entered has already been used in last N attempts"
 password.history.validation.error.1001=Password you entered has already been used in last N days"
```
Pushpalanka: Can we let the number N get populated with {} format?

PR #1215, opened 2017-04-05
IS 6 Admin Potal - Adding add group UI fixes

Minoli:


Minoli:
Fixed with PR https://github.com/wso2/product-is/pull/1216

PR #1246, opened 2017-05-19
Minor improvements to sample.

Pushpalanka:


modules/samples/oauth2/custom-grant/src/main/java/org/wso2/sample/identity/oauth2/grant/mobile/MobileGrant.java
```diff
@@ -154,4 +156,8 @@ private boolean isValidMobileNumber(String mobileNumber){
         return false;
     }
 
+    public boolean isOfTypeApplicationUser() throws IdentityOAuth2Exception {
```
pulasthi7: Do we need to make this method public?

Pushpalanka: This is an overridden method which has 'public' access modifier in super class.

modules/samples/oauth2/custom-grant/src/main/java/org/wso2/sample/identity/oauth2/grant/mobile/MobileGrant.java
```diff
@@ -154,4 +156,9 @@ private boolean isValidMobileNumber(String mobileNumber){
         return false;
     }
 
+    @Override
+    public boolean isOfTypeApplicationUser() throws IdentityOAuth2Exception {
+        return true;
```
pulasthi7: Do we need that as the super class returns true by default?

Pushpalanka: Is there any harm keeping this?  Even if superclass later change this value, it won't effect as we are explicitly setting it here and in one look the details are visible within the class if we keep it. Anyway I can remove if there is a thorough objection to keep it.. :)

PR #1278, opened 2017-07-17
Identity 3704

nilasini:
Added the test for unsupported grant type


nilasini:
this PR has been changed to [https://github.com/wso2/product-is/pull/1279]

PR #1282, opened 2017-07-27
Upgrading dependency versions.

rushmin:


darshanasbg:
These versions already updated as of https://github.com/wso2/product-is/blob/356cb0c5f4a8676163a5ef5f5b8f808163b82269/pom.xml. Hence closing the PR

PR #1305, opened 2017-08-14
 IDENTITY-6223:Change order of method calling in order to introduce decryption mechanism

nilasini:


pulasthi7:
As per the discussion with @nilasini, this won't work with existing agents. hence blocking the merge until we come to a resolution.

JKAUSHALYA:
As per the comment, closing this PR. Put a new PR with correct approach.

PR #1321, opened 2017-08-28
kernel version update

nuwandiw:


nuwandiw:
Do not merge. There are test failures to fix

PR #1344, opened 2017-08-31
removing informix db script

nuwandiw:


darshanasbg:
Seems like we don't have informix published as tested db in https://docs.wso2.com/display/compatibility/Tested+DBMSs.

darshanasbg:
Ref: https://github.com/wso2/carbon-identity-framework/pull/1036

PR #1363, opened 2017-09-08
Adding test cases for OAuth 2.0 DCRM create, read and delete requests

madurangasiriwardena:


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/oauth2/OAuth2ServiceAuthCodeGrantOpenIdTestCase.java
```diff
@@ -369,7 +369,10 @@ private void changeISConfiguration() throws Exception {
     private void resetISConfiguration() throws Exception {
 
         log.info("Replacing identity.xml with default configurations");
-        serverConfigurationManager.restoreToLastConfiguration();
+        File defaultIdentityXml = new File(getISResourceLocation() + File.separator + "default-identity.xml");
```
denuwanthi: Can't we move this and accompanying lines to a util class? It seems like this is repeated in several test cases

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/oauth2/dcrm/api/OAuthDCRMReadTestCase.java
```diff
@@ -0,0 +1,90 @@
...
+    @Test(alwaysRun = true, description = "Read request with an invalid client ID")
+    public void testReadServiceProviderWithInvalidClientID() throws IOException {
+        HttpGet request = new HttpGet(OAuthDCRMConstants.CLIENT_CONFIGURATION_ENDPOINT
+                + OAuthDCRMConstants.INVALID_CLIENT_ID);
+        request.addHeader(HttpHeaders.AUTHORIZATION, OAuthDCRMConstants.AUTHORIZATION);
+        request.addHeader(HttpHeaders.CONTENT_TYPE, OAuthDCRMConstants.CONTENT_TYPE);
+        request.addHeader(HttpHeaders.ACCEPT, OAuthDCRMConstants.ACCEPT);
+
+        HttpResponse response = client.execute(request);
+        assertEquals(response.getStatusLine().getStatusCode(), 500, "Service Provider read request " +
```
pulasthi7: According to the spec this should have returned a 401. So the implementation may be wrong to respond with 500 in this case. Hence correct the implementation and test case to reflect the spec.

In general, You should not assert a HTTP 500 unless there is a very special case. HTTP 500 can occur due to any reason and asserting a condition based on that is wrong.  This was done in several places, please check and correct.

madurangasiriwardena: Created https://wso2.org/jira/browse/IDENTITY-6379 to track the issue. Once the implementation is properly fixed, we will fix the test cases.

PR #1130, opened 2017-02-16
Implementation of Username Recovery feature.

GDRDABARERA:

PR #1371, opened 2017-09-13
Resolve PR comments of #1130

GDRDABARERA:


portal/apps/org.wso2.carbon.iam.user-portal.app/src/main/modules/recovery-manager.js
```diff
@@ -62,7 +62,8 @@ var recoveryManager = {};
     }
 
     /**
-     *  Check whether password recovery enabled.
+     * Check whether the password recovery enabled
```
kasunbg: The javadoc comments need to end with a dot. Otherwise, build fails in windows environment because of checkstyle.

portal/apps/org.wso2.carbon.iam.user-portal.app/src/main/modules/recovery-manager.js
```diff
@@ -62,7 +62,8 @@ var recoveryManager = {};
...
+     * @param
```
kasunbg: What is the purpose of adding this here?

PR #1385, opened 2017-09-28
IDENTITY-6377:Add db2 changes to the migration script

nilasini:
Add db2 changes to the migration script


nilasini:
This PR has some 5.4.0 script changes also. That must be removed since this is 5.2.0 to 5.3.0 migration script. Removed the 5.4.0 changes with https://github.com/wso2/product-is/pull/1386

PR #1418, opened 2017-10-23
IDENTITY-6811: Invoke csrf token injecting js from my login sessions gadget

malithie:


JKAUSHALYA:
@malithie can you make sure this is merged ?

PR #1457, opened 2017-11-24
IDENTITY-6924: Adding SAML sign keystore test cases

darshanasbg:
Added two main tests to test SAML sign keystore,
- SAML IdP metadata testcase
- SAML federation test case

Federation test case test two scenarios,
- SAML response & assertion signing (in inbound saml flow)
- SAML request signing (in the outbound saml flow)


JKAUSHALYA:
Merged with https://github.com/wso2/product-is/pull/1464

PR #1521, opened 2017-12-13
[IDENTITY-7034] Added variables to auto-fill the version, month, year.

ruwanta:
Added variables so that common changing properties are taken from variables.

### DIFF
--- a/modules/distribution/pom.xml
+++ b/modules/distribution/pom.xml
@@ -454,6 +454,36 @@
                     <skip>true</skip>
                 </configuration>
             </plugin>
+            <plugin>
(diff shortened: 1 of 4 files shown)


modules/distribution/pom.xml
```diff
@@ -454,6 +454,36 @@
                     <skip>true</skip>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>3.0.0</version>
```
madurangasiriwardena: shall we add this to pluginManagement in the root pom and remove the version from here?

PR #2046, opened 2018-01-11
Removing unnecessary files

madurangasiriwardena:
Below files are removed from the bin directory of the product distribution.

- java2wsdl.sh
- java2wsdl.bat
- wsdl2java.bat
- wsdl2java.sh
- tcpmon-1.0.jar
- tcpmon.sh
- tcpmon.bat

### DIFF
--- a/modules/distribution/src/assembly/bin.xml
+++ b/modules/distribution/src/assembly/bin.xml
@@ -40,11 +40,18 @@
                 <exclude>**/LICENSE.txt</exclude>
                 <exclude>**/release-notes.html</exclude>
                 <exclude>**/claim-config.xml</exclude>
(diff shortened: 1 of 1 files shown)


modules/distribution/src/assembly/bin.xml
```diff
@@ -40,11 +40,18 @@
                 <exclude>**/LICENSE.txt</exclude>
                 <exclude>**/release-notes.html</exclude>
                 <exclude>**/claim-config.xml</exclude>
-		<exclude>**/tomcat/catalina-server.xml</exclude>
+		        <exclude>**/tomcat/catalina-server.xml</exclude>
```
omindu: Fix formatting.

Issue #2260, opened 2018-02-07
Migration procedure from IS 5.3.0 to IS 5.4.0 - Migration does not work as expected


sagara-gunathunga:

nilasini:
One issue is migration client is migrating the inactive tenants.

The second issue is if the particular identity server doesn't have a local claim which is used by another claim dialect then there will be exception while migrating. So in order to avoid that need to add all the local claims which have dependency with other claim dialects added through migration.

nilasini:
Needs to check Oauth migration for inactive tenants as well.

PR #2268, opened 2018-02-07
Ignore claim migration for inactive tenant

nilasini:
Resolves:https://github.com/wso2/product-is/issues/2260
Ignoring the migration for inactive tenant, add all the local claims which have dependency with other claim dialects in migration.

### DIFF
--- a/modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v540/migrator/ClaimDataMigrator.java
+++ b/modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v540/migrator/ClaimDataMigrator.java
@@ -94,7 +94,14 @@ public class ClaimDataMigrator extends Migrator {
 
             // Migrate other tenants.
             List<Tenant> tenants = Utility.getTenants();
+            List<Integer> inactiveTenants = Utility.getInactiveTenants();
+            boolean ignoreForInactiveTenants = isIgnoreForInactiveTenants();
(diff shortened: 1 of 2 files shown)


modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v540/migrator/ClaimDataMigrator.java
```diff
@@ -94,7 +94,14 @@ public void migrate() throws MigrationClientException {
 
             // Migrate other tenants.
             List<Tenant> tenants = Utility.getTenants();
+            List<Integer> inactiveTenants = Utility.getInactiveTenants();
+            boolean ignoreForInactiveTenants = isIgnoreForInactiveTenants();
             for (Tenant tenant : tenants) {
+                int tenantId = tenant.getId();
```
darshanasbg: Should we have this fix only for ClaimDataMigrator module? Aren't there any other modules needed this fix?

nilasini: Currently other module migration are not depend on the tenants. We can change the migration of other modules as an impovement.

omindu:
This will only ignore claim migration for inactive tenants, correct ?

nilasini:
Yes it will ignore the inactive tenant according to the configuration in migration_config.yaml file.

PR #2329, opened 2018-02-14
add sample XACML policy for OAuth2 scope validator

senthalan:


modules/distribution/conf/policies/scope_based_token_validation_policy_template.xml
```diff
@@ -0,0 +1,55 @@
+<Policy xmlns="urn:oasis:names:tc:xacml:3.0:core:schema:wd-17"
```
Megala21: Please add the license header

PR #2532, opened 2018-02-20
Fixing stored XSS vulnerabilities in user profile UI

thanujalk:


thanujalk:
@malithie yes. It only used to write to the UI.

PR #2633, opened 2018-02-23
Remove XACML based scope validator sample policy and insert policy mover

senthalan:


senthalan:
As the policy will be moved to the repository auth-addons along with the core implementation.

PR #2733, opened 2018-03-01
Fix #2732: Set consent expiry time relative to browser timezone

### LINKED ISSUE #2732: Consent expiry time should be milliseconds relative to browser timezone when set from the gadget

omindu:
Currently we set the timestamp `Date.UTC(date picker date at 00:00)`. This gives the time in milliseconds for UTC date which is different than the browser time. Instead we should get the browser time in milliseconds.

### DIFF
--- a/modules/jaggery-apps/portal/gadgets/consent_management/js/gadget.js
+++ b/modules/jaggery-apps/portal/gadgets/consent_management/js/gadget.js
@@ -334,7 +334,7 @@ function addActions(container) {
         changeYear: true,
         onSelect: function (dateText) {
             var split = dateText.split("/");
-            var newDate = Date.UTC(split[2], split[0], split[1]);
+            var newDate = new Date(split[2], split[0] - 1, split[1]).getTime();
             var expiry = EXPIRY_DATE_STRING + newDate;
 
             $("#date_picker_new_expiry").val(expiry);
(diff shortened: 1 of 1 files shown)


modules/jaggery-apps/portal/gadgets/consent_management/js/gadget.js
```diff
@@ -503,7 +504,7 @@ function constructDate(expiry) {
     if (!checkValidDate(date)) {
         return getDefaultExpiry();
     }
-    var goodDate = date.getMonth() + "/" + date.getDate() + "/" + date.getFullYear();
+    var goodDate = date.getMonth() + 1 + "/" + date.getDate() + "/" + date.getFullYear();
```
hasinthaindrajee: Don't we have a way to construct the date directly without appending each part ? date format ?

omindu: @ayshsandu @thusithak Thoughts ?

thusithak: AFAIK there aren't any other straightforward way with vanilla js to achieve the same.
We could use some js library like [moment.js](https://momentjs.com/), which will be an overhead since we might not be doing alot of date format and manipulations.

Issue #2694, opened 2018-02-27
Update carbon dependencies in the product


emswbandara:
Need to update other carbon dependencies eg. carbon.registry versions in the product

PR #2752, opened 2018-03-04
Update carbon component versions

emswbandara:
Fixes https://github.com/wso2/product-is/issues/2694


modules/p2-profile-gen/pom.xml
```diff
@@ -429,6 +429,10 @@
                                 <featureArtifactDef>
                                     org.wso2.carbon.consent.mgt:org.wso2.carbon.consent.mgt.server.feature:${carbon.consent.mgt.version}
                                 </featureArtifactDef>
+
```
hasinthaindrajee: Do we need this new line ?

PR #2772, opened 2018-03-06
Denuwanthi syslog migration

denuwanthi:


modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v550/migrator/SysLogPropertiesMigrator.java
```diff
@@ -0,0 +1,56 @@
...
+import org.wso2.carbon.is.migration.service.Migrator;
+import org.wso2.carbon.is.migration.service.v550.RegistryDataManager;
+import org.wso2.carbon.is.migration.util.Constant;
+import org.wso2.carbon.registry.core.exceptions.RegistryException;
+import org.wso2.carbon.user.api.UserStoreException;
+
+public class SysLogPropertiesMigrator extends Migrator {
+
+    private static final Log log = LogFactory
+            .getLog(org.wso2.carbon.is.migration.service.v550.migrator.SysLogPropertiesMigrator.class);
```
nuwandiw: Do we need the class FQN here?

modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v550/migrator/SysLogPropertiesMigrator.java
```diff
@@ -0,0 +1,56 @@
...
+    public void migrate() throws MigrationClientException {
+
+        migrateSysLogPropertiesPassword();
+    }
+
+    private void migrateSysLogPropertiesPassword() throws MigrationClientException {
+
+        log.info(Constant.MIGRATION_LOG + "Migration starting on SYSLOG_PROPERTIES file");
+        try {
+            RegistryDataManager.getInstance().migrateSysLogPropertyPassword(true);
```
nuwandiw: Instead of setting value "true", use isIgnoreForInactiveTenants() method from Migrate super class so it will be configurable to skip inactive tenants.

PR #2881, opened 2018-03-10
Fix consents not updating after warning/confirmation message improvements

ayshsandu:
To update consents from user dashboard

### DIFF
--- a/modules/jaggery-apps/portal/gadgets/consent_management/js/gadget.js
+++ b/modules/jaggery-apps/portal/gadgets/consent_management/js/gadget.js
@@ -296,6 +296,7 @@ function renderReceiptDetails(data) {
     });
 
     container.bind('ready.jstree', function (event, data) {
+        allAttributes = [];
         var $tree = $(this);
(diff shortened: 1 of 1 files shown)


modules/jaggery-apps/portal/gadgets/consent_management/js/gadget.js
```diff
@@ -385,18 +386,18 @@ function addActions(container) {
     });
 }
 
-function showWarning() {
+function showWarning(container) {
```
pulasthi7: what's the usage of container here?

ayshsandu: container is an element need to be passed in the callBack function of the warning/confirmation dialog

pulasthi7: But the callback function is empty here. It's used in showConfirm()

ayshsandu: Fixed

PR #2979, opened 2018-03-17
Update next development versions to 5.6.0-SNAPSHOT

emswbandara:


omindu:
Do we need to change this to 5.6.0-SNAPSHOT ? Until 5.6.0 GA we should keep releasing 5.5.0-UPDATE-X and then change it to 5.6.0-SNAPSHOT after 5.6.0-GA. WDYT ?

darshanasbg:
@omindu @emswbandara We could keep the name as 5.6.0-SNAPSHOT as the the master..

Infact the version given in the master does not matter IMO. It could be 5.5.1-SNAPSHOT or 5.6.0-SNAPSHOT or 6.0.0-SNAPSHOT..

LTS version number gets decided from the PMM with the available features included in the product at a given point.

pulasthi7:
@darshanasbg @omindu @emswbandara 

+1 with omindu. We can have the name as 5.5.0-UPDATE-X-SNAPSHOT (X is the next update number), where we don't need to worry which is the next LTS version (for now)

Anyway, we'll need to do it soon, as the jenkins nightly builds will start failing due to 5.5.0 artifacts are already released to nexus.

PR #2986, opened 2018-03-21
Fix migration client to use correct snakeyaml version

nuwandiw:


modules/migration/migration-service/pom.xml
```diff
@@ -124,6 +124,7 @@
                             org.osgi.service.component; version="${osgi.service.component.imp.pkg.version.range}",
                             org.apache.commons.logging; version="${version.commons.logging}",
                             org.wso2.carbon.identity.core.*; version="${carbon.identity.framework.version.range}",
+                            snakeyaml; version="${snakeyaml.version}",
```
pulasthi7: trailing , is redundant

PR #3116, opened 2018-04-26
Implement sample to demonstrate MP-JWT feature

madawas:
This commit adds a sample service written using eclipse miroprofile 1.2 framework which is secured using mp-jwt 1.0.

Identity server can be used to generate a JWT token which can be used to invoke the endpoints of this service.

### DIFF
--- a/modules/samples/microprofile/microprofile-jwt/pom.xml
+++ b/modules/samples/microprofile/microprofile-jwt/pom.xml
@@ -17,7 +17,6 @@
         <dependency>
             <groupId>org.eclipse.microprofile</groupId>
             <artifactId>microprofile</artifactId>
-            <version>1.2</version>
             <scope>provided</scope>
             <type>pom</type>
         </dependency>
@@ -35,7 +34,6 @@
             <plugin>
                 <groupId>net.wasdev.wlp.maven.plugins</groupId>
                 <artifactId>liberty-maven-plugin</artifactId>
-                <version>2.2</version>
                 <configuration>
                     <assemblyArtifact>
                         <groupId>io.openliberty</groupId>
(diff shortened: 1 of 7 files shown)


modules/samples/microprofile/microprofile-jwt/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+                <artifactId>maven-war-plugin</artifactId>
+                <configuration>
+                    <failOnMissingWebXml>false</failOnMissingWebXml>
+                    <packagingExcludes>pom.xml</packagingExcludes>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>net.wasdev.wlp.maven.plugins</groupId>
+                <artifactId>liberty-maven-plugin</artifactId>
+                <version>2.2</version>
```
pulasthi7: Move the versions to the plugin management section in root pom

modules/samples/microprofile/microprofile-jwt/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+
+    <artifactId>microprofile-jwt</artifactId>
+
+    <packaging>war</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.microprofile</groupId>
+            <artifactId>microprofile</artifactId>
+            <version>1.2</version>
```
pulasthi7: Add this to dependency management section in root pom and remove the version,type from here

pom.xml
```diff
@@ -532,6 +532,11 @@
                     <artifactId>build-helper-maven-plugin</artifactId>
                     <version>3.0.0</version>
                 </plugin>
+                <plugin>
+                    <groupId>net.wasdev.wlp.maven.plugins</groupId>
+                    <artifactId>liberty-maven-plugin</artifactId>
+                    <version>2.2</version>
```
pulasthi7: Shall we define a property for this one also?

madawas: fixed with 7340ce6

modules/samples/microprofile/microprofile-jwt/pom.xml
madawas: Fixed with a3ae797

modules/samples/microprofile/microprofile-jwt/pom.xml
madawas: Fixed with a3ae797

PR #3131, opened 2018-04-30
Excluding liberty-plugin execution from the build

madawas:
Possible fix for the build failure.

Removing the execution part of the maven liberty plugin as maven-3.0.5 does not support Eclipse Aether. Microprofile JWT sample uses liberty maven plugin which requires Eclipse Aether.

https://wiki.eclipse.org/Aether/Using_Aether_in_Maven_Plugins


pulasthi7:
Let's also try to get the jenkins' maven version updated and add this back.

PR #3237, opened 2018-05-28
Migration client for 5.6.0

senthalan:
Migration for
- UMA and identity schema change
- Add new claims to active tenants (claims with MappedLocalClaim change are not automated)


modules/migration/migration-resources/5.6.0/dbscripts/step1/identity/h2.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE IDN_OAUTH_CONSUMER_APPS ADD ID_TOKEN_EXPIRE_TIME BIGINT DEFAULT 3600000;
+
+CREATE TABLE IDN_AUTH_SESSION_STORE_TEMP (
```
nuwandiw: As per an internal discussion, the table name would be changed to IDN_AUTH_TEMP_SESSION_DATA_STORE. The scripts should be changed.

Also a new index is added to IDN_AUTH_SESSION_STORE table. Similar index should be used in TEMP table as well. Contact @rushmin on this matter.

modules/migration/migration-resources/migration-config.yaml
```diff
@@ -194,3 +194,21 @@ versions:
...
+        schema: "uma"
+    -
+      name: "SchemaMigrator"
+      order: 2
+      parameters:
+        location: "step1"
+        schema: "identity"
+    -
+      name: "ClaimDataMigrator"
+      order: 3
```
nuwandiw: Add new line at the end of the file

modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v560/migrator/ClaimDataMigrator.java
```diff
@@ -0,0 +1,293 @@
...
+     * Migrate claim mappings.
+     *
+     * @param tenantId tenant id
+     * @throws UserStoreException     UserStoreException
+     * @throws ClaimMetadataException ClaimMetadataException
+     */
+    private void migrateClaimData(int tenantId) throws UserStoreException, ClaimMetadataException {
+
+        UserRealm realm = ISMigrationServiceDataHolder.getRealmService().getTenantUserRealm(tenantId);
+        String primaryDomainName = realm.getRealmConfiguration().getUserStoreProperty(UserCoreConstants.RealmConfig
```
nuwandiw: Is there a possibility of realm attribute having a null value for an invalid tenant id? If so handle null case to avoid NPE.

modules/migration/migration-resources/5.6.0/dbscripts/step1/identity/h2.sql
senthalan: I will contact with @rushmin  and update the script

modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v560/migrator/ClaimDataMigrator.java
```diff
@@ -0,0 +1,296 @@
...
+ */
+public class ClaimDataMigrator extends Migrator {
+
+    private static final Log log = LogFactory.getLog(ClaimDataMigrator.class);
+
+    private static final String CLAIM_CONFIG = "claim-config.xml";
+
+    private ClaimConfig claimConfig;
+
+    private ClaimDialectDAO claimDialectDAO = new ClaimDialectDAO();
```
nuwandiw: Since claimDialectDAO variable is only used in migrateClaimData method, shall we define a local variable in the method instead of this instance variable

modules/migration/migration-resources/migration-config.yaml
senthalan: added

modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v560/migrator/ClaimDataMigrator.java
```diff
@@ -0,0 +1,296 @@
...
+        } catch (IOException | XMLStreamException | UserStoreException e) {
+            String message = "Error while building claims from config file";
+            if (isContinueOnError()) {
+                log.error(message, e);
+            } else {
+                throw new MigrationClientException(message, e);
+            }
+        }
+
+        if (claimConfig.getClaims().isEmpty()) {
```
nuwandiw: If this line is executed after Line#84, claimConfig  would not be initialized properly. Better to move this part in to try block.

modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v560/migrator/ClaimDataMigrator.java
```diff
@@ -0,0 +1,296 @@
...
+
+        Map<String, String> claimProperties = claimConfig.getPropertyHolder().get(claimURI);
+        claimProperties.remove(ClaimConstants.DIALECT_PROPERTY);
+        claimProperties.remove(ClaimConstants.CLAIM_URI_PROPERTY);
+        claimProperties.remove(ClaimConstants.ATTRIBUTE_ID_PROPERTY);
+
+        if (!claimProperties.containsKey(ClaimConstants.DISPLAY_NAME_PROPERTY)) {
+            claimProperties.put(ClaimConstants.DISPLAY_NAME_PROPERTY, "0");
+        }
+        if (claimProperties.containsKey(ClaimConstants.SUPPORTED_BY_DEFAULT_PROPERTY) &&
```
nuwandiw: Is this the correct logic for setting supported_by_default, read_only and required to true? Check and verify

senthalan: I checked with the DefaultClaimMetadataStore implementation. This same logic is used

PR #3521, opened 2018-08-03
Migration client fixes for oracle db

senthalan:
Migration client fixes for oracle db when migrating from 5.2.0 to 5.6.0


denuwanthi:
@senthalan , so this means we have checked direct migration from 5.2 to 5.6 on oracle DB?

senthalan:
@denuwanthi, yes. And we have planned to check direct migrations with other databases also in the future.

PR #3551, opened 2018-08-09
Add tenant range migration option for IS 5.4.0 and IS 5.5.0 migration clients.

kasun-bandara:
This will add the capability to migrate the selected tenant by specifying following in the "migration-config.yaml"

migrateTenantRange: "true"
migrationStartingTenantID: "100"
migrationEndingTenantID: "200"

Note: This fix is only for the IS 5.4.0 and IS 5.5.0 migration clients.


senthalan:
Do we need to implement this range option to other clients as well?

PR #3772, opened 2018-09-21
Fix migration client

nuwandiw:


modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v550/migrator/OAuthDataMigrator.java
```diff
@@ -63,8 +63,12 @@ public void migrate() throws MigrationClientException {
     public void addHashColumns() throws MigrationClientException, SQLException {
 
         try (Connection connection = getDataSource().getConnection()) {
-            isTokenHashColumnsAvailable = TokenDAO.getInstance().isTokenHashColumnsAvailable(connection);
-            isAuthzCodeHashColumnAvailable = AuthzCodeDAO.getInstance().isAuthzCodeHashColumnAvailable(connection);
+            try {
+                isTokenHashColumnsAvailable = TokenDAO.getInstance().isTokenHashColumnsAvailable(connection);
+                isAuthzCodeHashColumnAvailable = AuthzCodeDAO.getInstance().isAuthzCodeHashColumnAvailable(connection);
+            } finally {
+                connection.commit();
```
denuwanthi: When using try with resources block, connection commit is auto handled I suppose. Then no need to specifically commit again. Please check on that

nuwandiw: When using try with resource, it will automatically close the connection. Connection commit is not handled by it. Oracle documentation can be found in [1]. Also 1st answer in [2] explains how connection treated within try-with-resource.

[1] https://docs.oracle.com/javase/8/docs/technotes/guides/language/try-with-resources.html
[2] https://stackoverflow.com/questions/15761791/transaction-rollback-on-sqlexception-using-new-try-with-resources-block

PR #3794, opened 2018-10-02
Adding token hashing migration

nuwandiw:


denuwanthi:
We might need to do the same for authorization code hash also I think.

modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v570/dao/TokenDAO.java
```diff
@@ -0,0 +1,57 @@
...
+        return oauthTokenInfos;
+    }
+
+    public void updateNewTokenHash(List<OauthTokenInfo> updatedOauthTokenList, Connection connection)
+            throws SQLException {
+
+        try (PreparedStatement preparedStatement = connection.prepareStatement(UPDATE_ACCESS_TOKEN)) {
+            for (OauthTokenInfo oauthTokenInfo : updatedOauthTokenList) {
+                preparedStatement.setString(1, oauthTokenInfo.getAccessTokenHash());
+                preparedStatement.setString(2, oauthTokenInfo.getAccessTokenHash());
```
denuwanthi: shouldn't this be oauthtokeninfo.rereshTokenhash?

PR #3796, opened 2018-10-03
Add migration client fixes for tenant range option and permission migration

kasun-bandara:
This PR will,

- Add Tenant range migration improvements
- Resolve the connection leak issue in permission migration


denuwanthi:
Thank you for the contribution @kasun-bandara

PR #3946, opened 2018-11-09
Test scenarios for user provisioning using Admin Services

shavantha:
Test scenarios for user provisioning using Admin Services: 
Created by:shavantha weerasinghe


shavantha:
Based on the discussion with the IS team and STL team the scenarios may change so I will close this PR

PR #3962, opened 2018-11-13
Modify directory structure

janakamarasena:
Modify directory structure


product-scenarios/pom.xml
```diff
@@ -20,6 +20,7 @@
         <groupId>org.wso2</groupId>
         <artifactId>wso2</artifactId>
         <version>1</version>
+        <relativePath></relativePath>
```
omindu: Is this required ?

janakamarasena: It is required else maven takes IS parent as the parent pom.
https://stackoverflow.com/questions/6003831/parent-relativepath-points-at-my-com-mycompanymyproject-instead-of-org-apache

PR #3963, opened 2018-11-13
Update identity.outbound.auth.oidc.version

deki:
Update identity.outbound.auth.oidc.version to include fixes in next IS release


isharailanga:
Fixed in https://github.com/wso2/product-is/pull/3978, hence we can close this PR.

PR #4028, opened 2018-11-23
updated  the All attributes provisioning  test case to use the SCIM Util class for provisioning

shavantha:
The test case was updated to call the util class for scim. the test case looks at using all attributes when provisioning a new user.

tested environments
1.ubuntu 18.04
2.java 1.8.0_161


shavantha:
closing the PR until the cleanup

PR #4053, opened 2018-11-28
The SCIM2.0 test cases cover validation in the use of proper content type

shavantha:
Purpose
the test cases validate if the system will provide the required response codes if we skip sending the content type or pass a wrong content type

Approach
1.This test case sends a wrong content type when provisioning a user

Tested environment 
1.ubuntu 18.04
2.java version "1.8.0_161"


shavantha:

PR #4057, opened 2018-11-28
Provision with all attributes scim2 test case

shavantha:
Purpose
This test case will provision a user with all attributes and assert for a 201 response code

Approach
This test case sends all attributes that can provision a user but not the full request

Tested environments
1.ubuntu 18.04
2.java version "1.8.0_161"
3.Identity Server fresh pack with embedded ldap and Jdbc Userstore


shavantha:

PR #4078, opened 2018-11-30
The test case calls the Me endpoint with authentication header

shavantha:
Purpose
This test case will  call the scim2/Me endpoint to self register a user 

Approach
This test case self register a user calling the scim2/Me endpoint and then asserts the user name. Next, the cleanup method will be called to delete the created user. The deletion is not through the scim2/Me endpoint but by passing the userId.

Tested environments
1.ubuntu 18.04
2.java version "1.8.0_161"
3.Identity Server with  JDBC Userstore and mysql database


shavantha:

PR #4083, opened 2018-11-30
The test case submits a user as an email format when jdbc user store is used with default regex

shavantha:
Purpose
This test case submits a user as an email format when jdbc user store is used with default regex

Approach
This test case can be run when jdbc user store is configured with the default regex. In this situation the user name will get creted as ex: user@mail.com as the below pattern allows.

 <Property name="UsernameJavaRegEx">^[\S]{3,30}$</Property>
<Property name="UsernameJavaScriptRegEx">^[\S]{3,30}$</Property>

Tested environments
1.ubuntu 18.04
2.java version "1.8.0_161"
3.Identity Server with JDBC Userstore and mysql database


shavantha:

PR #4109, opened 2018-12-03
Test cases that validate the content type

shavantha:
Purpose
the test cases validate if the system will provide the required response codes if we skip sending the content type or pass a wrong content type

Approach
1.This test case sends a wrong content type when provisioning a user

Tested environment
1.ubuntu 18.04
2.java version "1.8.0_161"


shavantha:
closing this PR as a new PR https://github.com/wso2/product-is/pull/4154 is created

PR #4112, opened 2018-12-03
The test case validates for different content type in scim 1.1 api

shavantha:
Purpose
This test case sends checks if wrong content types are validated

Tested environment
1.ubuntu 18.04
2.java version "1.8.0_161"
3.Identity server wum updated pack wso2is-5.7.0+1543336788716.full with jdbc user store

Automation tests
The test attempts to provision a user with application/xml as the content type for scim 1.1 api and asserts the response code


shavantha:
closing this PR will be sending 1 PR for scim 1.1 scenarios

Issue #4022, opened 2018-11-23
SCIM1.1 Wrong error code is returned by the Identity Server for unauthorized /forbidden user scenario


shavantha:
Wrong error code is returned by the Identity Server for unauthorized user scenario

pack :wum updated
wso2is-5.7.0+1541435691728.full.zip

test environment
1.Ubuntu 18.04
2.Java 1.8.0_161

steps
1.with scim 1.1 api create a user with the admin credentials. ex:scim1user
2.using the newly created user scim1user attempt to create another user ex:scim2user
Note: The below error is shown when running the test

 User is not authorized to perform provisioning expected [401] but found [500]. Ideally the error code should be 401 and not 500

From the management console, if we perform the above steps 1 them for step 2 the user is not shown the add user options

testSCIMCreateSecondUser(org.wso2.identity.scenarios.test.scim.UserProvisionWithInsufficientPrivilegesSCIMTestCase)  Time elapsed: 0.134 sec  <<< FAILURE!
java.lang.AssertionError: User is not authorized to perform provisioning expected [401] but found [500]
	at org.testng.Assert.fail(Assert.java:94)
	at org.testng.Assert.failNotEquals(Assert.java:513)
	at org.testng.Assert.assertEqualsImpl(Assert.java:135)
	at org.testng.Assert.assertEquals(Assert.java:116)
	at org.testng.Assert.assertEquals(Assert.java:389)
	at org.wso2.identity.scenarios.test.scim.UserProvisionWithInsufficientPrivilegesSCIMTestCase.testSCIMCreateSecondUser(UserProvisionWithInsufficientPrivilegesSCIMTestCase.java:92)

shavantha:
even invoking with curl the error code comes as 500 ideally this should be 401
{"Errors":[{"code":"500","description":"User is not authorized to perform provisioning"}]}

shavantha:
Second scenario
1.provision a user in scim
2.assign the admin role to the user
3.attempt to delete self using the user created in step1 with the admin role assigned in step2

Note: instead of 403 the system returns a 500

isharak:
Thank you for your contribution!
We are closing this issue since it has not been prioritized for a long time. Chances are that it has already been solved in more recent versions. If not, we will be re-evaluating this when it becomes a priority.

PR #4113, opened 2018-12-03
SCIM1.1 This test case attempts to provision a user using unauthorize credentials

shavantha:
Purpose
This test case has the following scenario
1.The admin adds a user ex:(scim1user) using the scim1.1 API
2.We attempt to create a second user ex: (scim2user) using the credentials of scim1user
3.Since scim1user does not have adequate permission we assert for [401] User is not authorized to perform provisioning expected.

Known Issues
#4022

Tested environments
1.Ubuntu18.04
2.Java: 1.8.0_161


shavantha:
closing this PR will be sending 1 PR for scim 1.1 scenarios

PR #4115, opened 2018-12-03
Add migration scripts for user session management

Abilashini:
This includes the migration of DB scripts to create the newly introduced tables for user session management.

Used by: https://github.com/wso2/carbon-identity-framework/pull/2020


emswbandara:
@Abilashini can you resolve the conflicts in this PR?

modules/migration/migration-service/src/main/java/org/wso2/carbon/is/migration/service/v580/V580Migration.java
```diff
@@ -0,0 +1,32 @@
+/*
+ *   Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
nilasini: Change the year

nilasini:
@Abilashini can you please resolve the conflicts

senthalan:
Please resolve the config

modules/migration/migration-resources/5.8.0/dbscripts/step1/identity/postgresql.sql
```diff
@@ -1 +1,21 @@
...
+CREATE TABLE IDN_AUTH_USER (
+	USER_ID VARCHAR(255) NOT NULL,
+	USER_NAME VARCHAR(255) NOT NULL,
+	TENANT_ID INTEGER NOT NULL,
+	DOMAIN_NAME VARCHAR(255) NOT NULL,
+	IDP_ID INTEGER NOT NULL,
+	PRIMARY KEY (USER_ID),
+	CONSTRAINT USER_STORE_CONSTRAINT UNIQUE (USER_NAME, TENANT_ID, DOMAIN_NAME, IDP_ID));
+
+DROP TABLE IF EXISTS IDN_AUTH_USER_SESSION_MAPPING;
```
omindu: in which version this table was introduced ?

Abilashini: For IS 5.2.0 and 5.7.0 we have patched. For IS 5.8.0 it will be added by default. Tables are added at https://github.com/wso2/carbon-identity-framework/pull/2020

omindu: In that case dropping the tables is not correct. If the user already has this in older version, they'll lose all the data.

The other concern is, we have to check whether the table exist before adding it.

Abilashini: fixed

PR #4131, opened 2018-12-05
SCIM2 test case for provisioning a user with insufficient permission

shavantha:
Purpose
This test case validates for permission when attempting to provision a user

Tested environments
1.Ubuntu18.04
2.Java: 1.8.0_161

Test case
1.The test case initially provisions a user(ex:scim2user) by a user with admin credentials and then we try to create a second user by using scim2user who does not have adequate permissions. We assert for SC_FORBIDDEN status code


shavantha:
closing this PR will be sending this with another PR

PR #4137, opened 2018-12-06
SCIM2 Test case to update a provisioned users profile

shavantha:
Purpose
This test case updates a provisioned users profile attribute "first name" and the "last name" and assert if the first name is what we changed

Tested environment
1.ubuntu 18.04
2.java version "1.8.0_161"
3.Identity server wum updated pack wso2is-5.7.0+1543336788716.full with jdbc user store with the UsernameJavaRegEx changes to [a-zA-Z0-9._-|//]{3,30}$

Automation tests
The test attempts to update a provisioned users profile. First a user is provisioned and then the first name attribute is added with a value. Next with the updated users method we update the provisioned users profile firstName


shavantha:
closing this PR as the new PR is https://github.com/wso2/product-is/pull/4154

PR #4168, opened 2018-12-08
SCIM2 Testcases

shavantha:
Purpose
This PR contains multiple SCIM2,0 Test cases two of the test cases are to be run on a different combination.

Known Issues
N/A

Tested environments
1.Ubuntu18.04
2.Java: 1.8.0_161

Test cases
1.AnonymousProvisioningWithoutHeaderTestCase : run this test case only if    Resource context="(.*)/scim2/Me" secured="true" http-method="POST" is set to false

2.FilterOutEmailUserProvisioningSCIM2TestCase.java :  Run this test case only when the default value of UsernameJavaRegEx in JDBC user store has been changed  to block @ sign ex:[a-zA-Z0-9._\-|//]{3,30}$ . This test case asserts that the error schema is returned

3.UpdateProvisionedUserRoleSCIM2TestCase.java : This test case will create a new user and a user role and assign the new role "Engineering" to the scim2user. This test case then asserts that the new role name exist.


shavantha:
Closing this PR : Reason is that we have decided not to use the util class but to implement a base class for user provisioning

PR #4239, opened 2018-12-21
Improve the test execution time

pulasthi7:


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/provisioning/DBSeperationTestCase.java
```diff
@@ -26,9 +26,7 @@
 import org.wso2.carbon.identity.application.common.model.idp.xsd.IdentityProvider;
```
vihanga-liyanage: shall we fix the spelling mistake in the class name as well? Should be DBSep**a**rationTestCase.

PR #4411, opened 2019-02-05
Adding '2.1 User self registration via end user portal' test case.

thanujalk:


thanujalk:
Merging to get more details on the current live build failures.

PR #4449, opened 2019-02-15
Correct u2f-api.js url in u2f registration page

MohammedEssehemy:
handle wrong script url which cause 404 error in user_profile page

### DIFF
--- a/modules/jaggery-apps/portal/gadgets/user_profile/js/gadget.js
+++ b/modules/jaggery-apps/portal/gadgets/user_profile/js/gadget.js
@@ -1,6 +1,6 @@
 function drawPage() {
     var output = "";
-    var u2fScript="<script src=\"u2f-api.js\"></script>";
+    var u2fScript="<script src=\"/portal/gadgets/user_profile/js/u2f-api.js\"></script>";
     var start = "<div class=\"container-fluid\" style=\"width:95%\">\n" +
         "    <div class=\"row\">\n" +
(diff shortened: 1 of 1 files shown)


MohammedEssehemy:
any updates here?

modules/jaggery-apps/portal/gadgets/user_profile/js/gadget.js
```diff
@@ -1,6 +1,6 @@
 function drawPage() {
     var output = "";
-    var u2fScript="<script src=\"u2f-api.js\"></script>";
+    var u2fScript="<script src=\"/portal/gadgets/user_profile/js/u2f-api.js\"></script>";
```
ruwanta: The problem in adding url starting with "/" prevent the portal to be re-mapped to different context path via reverse-proxy.
The idea of having the relative URL based on the current page (i.e /portal/gadgets/user_profile/js/")  makes it available even the context path is remapped.
@MohammedEssehemy , can you explain what was to motive to think the original URL is incorrect.

MohammedEssehemy: To handle wrong script url which cause 404 error in browser console in user_profile page

IsurangaPerera:
From IS590 we'll be using WebAuthn API at the client-side of FIDO implementation. Furthermore, most browsers have dropped the support to U2F extension. 
Hence closing the issue.

@MohammedEssehemy appreciate your contributions to the project 👌

GDRDABARERA:
Closing as this is related to the old APIs

PR #4522, opened 2019-03-04
Add identity-event.properties templating.

hasinidilanka:
**Proposed Changes**

- Add `identity-event.properties` templating.
- Add default configurations to `default.json` file

- Read custom configurations in the template file.

- Example Toml configuration file for custom events

```
[[identity.custom_events.schemes]]
name="custom1"
subscriptions=["sub1","sub2"]
[identity.custom_events.schemes.properties]
enable=true
```

### DIFF
--- a/modules/distribution/src/repository/resources/conf/default.json
+++ b/modules/distribution/src/repository/resources/conf/default.json
@@ -235,5 +235,188 @@
     "org.wso2.carbon.identity.entitlement.pip.DefaultResourceFinder"
   ],
   "identity.entitlement.entitlement_engine_caching_interval": "100000",
-  "identity.entitlement.JSON_shorten_form_enabled": false
+  "identity.entitlement.JSON_shorten_form_enabled": false,
+
+
+  "identity.events.threadPool_size": "10",
+  "identity.events.schemes": {
+    "account.lock.handler": {
+      "subscriptions": [
+        "PRE_AUTHENTICATION",
+        "POST_AUTHENTICATION",
+        "PRE_SET_USER_CLAIMS",
+        "POST_SET_USER_CLAIMS"
+      ],
+      "properties": {
+        "enable": false,
+        "login.fail.timeout.ratio": "2",
+        "On.Failure.Max.Attempts": "5",
+        "Time": "5",
+        "notification.manageInternally": true
+      }
+    },
+    "emailSend": {
+      "subscriptions": [
+        "TRIGGER_NOTIFICATION"
+      ]
+    },
+    "accountConfirmationValidation": {
+      "subscriptions": [
+        "PRE_AUTHENTICATION"
+      ]
+    },
+    "passwordHistory": {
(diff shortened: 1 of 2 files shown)


modules/distribution/src/repository/resources/conf/default.json
```diff
@@ -235,5 +235,188 @@
     "org.wso2.carbon.identity.entitlement.pip.DefaultResourceFinder"
   ],
   "identity.entitlement.entitlement_engine_caching_interval": "100000",
-  "identity.entitlement.JSON_shorten_form_enabled": false
+  "identity.entitlement.JSON_shorten_form_enabled": false,
+
+
+  "identity.events.threadPool_size": "10",
```
pulasthi7: shall we have this as "thread_pool_size"

modules/distribution/src/repository/resources/conf/default.json
```diff
@@ -235,5 +235,188 @@
...
+      "subscriptions": [
+        "PRE_AUTHENTICATION",
+        "POST_AUTHENTICATION",
+        "PRE_SET_USER_CLAIMS",
+        "POST_SET_USER_CLAIMS"
+      ],
+      "properties": {
+        "enable": false,
+        "login.fail.timeout.ratio": "2",
+        "On.Failure.Max.Attempts": "5",
```
pulasthi7: shouldn't we enclose these in quotes?

hasinidilanka: When we enclose them in single quotes, single quotes also get printed in the jinja template file.

`account.lock.handler.'On.Failure.Max.Attempts'=5`

modules/distribution/src/repository/resources/conf/default.json
hasinidilanka: Fixed by [073dd55](https://github.com/wso2/product-is/pull/4522/commits/073dd558df49c3ce182e4ec6aa1b07367c22f0ec)

modules/distribution/src/repository/resources/conf/default.json
hasinidilanka: Fixed.

modules/distribution/src/repository/resources/conf/default.json
```diff
@@ -235,5 +235,127 @@
     "org.wso2.carbon.identity.entitlement.pip.DefaultResourceFinder"
   ],
   "identity.entitlement.entitlement_engine_caching_interval": "100000",
-  "identity.entitlement.JSON_shorten_form_enabled": false
+  "identity.entitlement.JSON_shorten_form_enabled": false,
```
denuwanthi: why can't we use lower case 'json' here?

hasinidilanka: Since it is a standard abbreviation I have used capital. If we need to fix we can introduce a key-mapping to maintain the backward compatibility.

PR #4524, opened 2019-03-05
Update component versions & fix test failures.

thanujalk:


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/oauth2/OIDCDiscoveryTestCase.java
```diff
@@ -104,12 +100,7 @@ public void testDiscovery() throws IOException {
         clientConfig.handlers(basicAuth);
 
         RestClient restClient = new RestClient(clientConfig);
-        String discoveryUrl;
-        if(discoveryBasePath.endsWith("/")){
-            discoveryUrl = discoveryBasePath + ".well-known/openid-configuration";
-        }else {
-            discoveryUrl = discoveryBasePath + "/.well-known/openid-configuration";
-        }
+        String discoveryUrl = isServerBackendUrl + "/oauth2/oidcdiscovery/.well-known/openid-configuration";
```
madurangasiriwardena: Shouldn't we be checking the url
`isServerBackendUrl + "/oauth2/token/.well-known/openid-configuration"`
also, which is the spec complient url?
The one we have checked here is not the correct url and its working to make it backward compatible only.

Issue #4434, opened 2019-02-13
Dashboard backend admin service URL is not configurable


rksk:
If we are setting up a proxy for the identity server and set proxy host and port as mentioned in [doc](https://docs.wso2.com/display/IS530/Setting+Up+Deployment+Pattern+1#SettingUpDeploymentPattern1-Changinghostnamesandports), then all the admin service calls from the dashboard and portal jaggery apps are using `https://proxy_host:proxyport/services` URL.

There are some cases, we need to avoid these admin service calls going through proxy and hit the backend direclty.

According to [doc](https://docs.wso2.com/display/IS530/Setting+Up+Deployment+Pattern+1#SettingUpDeploymentPattern1-Settingupthedashboard), it seems that URL is configurable using `site.json`. But according to the [code](https://github.com/wso2/product-is/blob/master/modules/jaggery-apps/user-dashboard/dashboard/util/utility.jag), the `proxyHost` and `proxyPort` set in `site.json` is discarded in the `setServerURL()`.

nilasini:
Fixed with 
https://github.com/wso2/product-is/pull/4518
https://github.com/wso2/product-is/pull/4662

PR #4662, opened 2019-03-19
Fixing some issues in configuring backend service Url in Dashboard

piraveena:


modules/jaggery-apps/portal/conf/site.json
```diff
@@ -1,6 +1,11 @@
 {
     "server" : {
```
malithie: I think we changed this to 'service'

malithie:
Link this to the issue and mention the approach followed

piraveena:
This is related to the issue [4434](https://github.com/wso2/product-is/issues/4434). 

When IS is fronted with LB, the dashboard works only if reverse proxy is configured in IS (reverse proxy can be configured by configuring hostname and proxy port in carbon.xml file and catalina-server.xml files respectively). 

With this fix,  dashboard can work without configuring carbon.xml and catalina-server.xml files. If reverse proxy is not configured in IS side, then we need to  configure host name and port in site.json files, SAMLIdpUrl , and sso-idp-config.xml

modules/jaggery-apps/portal/util/utility.jag
```diff
@@ -23,18 +23,60 @@ function setServerURL() {
...
         var serverCoreUrl = '';
 
-        var serviceOrigin = jsonConf.server.serviceOrigin;
+        var mgtTransport = "https";
         var fidoAppId = jsonConf.fido.appId;
+        //retrieving site configuration for overiding ui urls
+        var proxyConfHost = jsonConf.proxy.proxyHost;
+        var proxyConfPort = jsonConf.proxy.proxyHTTPSPort;
+        var proxyOrigin = '';
+        var proxyContextPath = ''
```
omindu: missing `;`

PR #4814, opened 2019-03-26
Templating jaas.conf file.

hasinidilanka:
**Proposed Changes**

- Templating jaas.conf file


hasinidilanka:
Seems this file doesn't need any modification for any scenarios. Only usage is for IWA kerberos and that doesn't need any modification from the shipped version. Hence we'll put this on hold now. And only merge if required.

PR #4877, opened 2019-03-29
Bump product-is dependency versions

DMHP:


pom.xml
```diff
@@ -2048,4 +2048,4 @@
     </scm>
 
 
-</project>
+</project>
```
mefarazath: New line please 😉

Issue #3744, opened 2018-09-12
Update the application name on consent screen when login to dashboard


Minoli:
The application name should be displayed to user without underscores. eg. WSO2 SP Dashboard

![dashboard-consent](https://user-images.githubusercontent.com/9837443/45415553-8a395780-b69b-11e8-9355-b424e4775957.png)

darshanasbg:
As mentioned by @madurangasiriwardena in https://github.com/wso2/product-is/pull/4912#issuecomment-560298871

> In the same time dashboard application is deprecated and will no longer be shipped with the product. 5.10.0 release will contain a brand new user portal application that will contain similar name as you have fixed.

The generic issue of showing proper display name for any application during the consent page is raised with: https://github.com/wso2/product-is/issues/6950.

Hence closing this issue.

PR #4912, opened 2019-04-02
Update the application name on consent screen when login to dashboard

SwatiNEC:
Update the application name on consent screen when login to dashboard [#3744](https://github.com/wso2/product-is/issues/3744)

### DIFF
--- a/modules/jaggery-apps/user-dashboard/sp_dashboard.xml
+++ b/modules/jaggery-apps/user-dashboard/sp_dashboard.xml
@@ -1,6 +1,6 @@
 <ServiceProvider>
 	<ApplicationID>2</ApplicationID>
-	<ApplicationName>wso2_sp_dashboard</ApplicationName>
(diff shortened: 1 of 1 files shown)


darshanasbg:
@thanujalk : Ain't we planning to add seperate display name for Service Provider to use places like this?

thanujalk:
> @thanujalk : Ain't we planning to add seperate display name for Service Provider to use places like this?

@darshanasbg : We are allowing spaces in the service provider name. So the same value can be used as the display name.

darshanasbg:
The generic issue tracked with: https://github.com/wso2/product-is/pull/4912

madurangasiriwardena:
@SwatiNEC, Thank you for your contribution. This file is no longer in this place. PR https://github.com/wso2/product-is/pull/6037 has restructured this file and the module. So this PR has conflicts.

In the same time dashboard application is deprecated and will no longer be shipped with the product. 5.10.0 release will contain a brand new user portal application that will contain similar name as you have fixed.

Issue #4440, opened 2019-02-14
Mapping multiple roles into one from ADFS IDP to WSO2 results in broken JWT on first login


kuche1991:
WSOIS-Version: **5.7**

The situation is, that we use wso2is with an ADFS backend as identity provider. We now wanted to map multiple roles of the ADFS-user into one role on wso-side so that, if one of the roles is present in ADFS the role in wso will be set.

User-Role in ADFS | Rolemapping in WSO2 IDP Settings | What's happening
------------ | ------------- | --------------
test | test -> admin, admin -> admin | works, provisioned user gets role admin
admin | test -> admin, admin -> admin | works, provisioned user gets role admin
admin, test | test -> admin, admin -> admin | If the user doesn't exist, the first login-try will result in a broken JWT-Token without any roles in it and the WSO-log-error: `Error occurred while updating hybrid role list of user`. Now for the strange part: If you login again, everything works according to the assumption and the user gets the admin role. This also happens if the user exists without the role, you change the roles on ADFS-side and re-login. The first time you get the error and the user in WSO won't get assigned to the role. If you do the same login again, it works and the role is present in the user. This feels really strange to me.

I hope you can help me :)

dilin993:
We were able to reproduce a similar issue, while trying out the following flow.
-  Configure two Identity Servers(IS) with OAUTH2 such that one IS acts as an federated IdP to the other
-  Create a user and assign "admin" and "test" roles at the IS acting as IdP
-  Add role mappings to the Identity Provider in the other IS (test -> admin, admin -> admin)
- Enable JIT provisioning

After this try to login to an application in the primary IS using the user created above. When the user get provisioned the following error trace is displayed. 
```
[2019-04-05 11:27:30,853] ERROR {org.wso2.carbon.identity.application.authentication.framework.handler.sequence.impl.DefaultStepBasedSequenceHandler} -  User provisioning failed!
org.wso2.carbon.identity.application.authentication.framework.exception.FrameworkException: Error while provisioning user : userD
	at org.wso2.carbon.identity.application.authentication.framework.handler.provisioning.impl.DefaultProvisioningHandler.handle(DefaultProvisioningHandler.java:181)
	at org.wso2.carbon.identity.application.authentication.framework.handler.sequence.impl.DefaultStepBasedSequenceHandler.handleJitProvisioning(DefaultStepBasedSequenceHandler.java:610)
[...]

PR #4949, opened 2019-04-04
Bump ldap and framework versions

dilin993:
https://github.com/wso2/product-is/issues/4440


malithie:
Addressed with https://github.com/wso2/product-is/pull/4951

Issue #4927, opened 2019-04-03
Update rampart version in the product


madurangasiriwardena:
Update rampart version in the product to fix

- https://github.com/wso2/product-is/issues/4571

PR #4959, opened 2019-04-05
Update versions in rampart and charon in product-is

DMHP:


DMHP:
resolves https://github.com/wso2/product-is/issues/4927

PR #5062, opened 2019-04-16
Add 'enable enchance filtering' config for SCIM.

hasinidilanka:
**Proposed Changes**

- Template `EnableFilteringEnhancements` in identity.xml.

- This config is mainly used to filter users in the primary domain.


modules/distribution/src/repository/resources/conf/default.json
```diff
@@ -578,6 +578,7 @@
   "scim2.endpoints.users_endpoint": "$ref{server.base_path}/wso2/scim2/Users",
   "scim2.endpoints.groups_endpoint": "$ref{server.base_path}/wso2/scim2/Groups",
   "scim2.enable_complex_multivalued_attribute_support": false,
+  "scim2.enable_enhance_filtering" : false,
```
pulasthi7: Shouldn't this be true for new releases?
And let's have the name 'enable_filtering_enhancements' or something similar. (Avoid double verbs)

hasinidilanka:
This is moved to the component level hence closing the PR.

Issue #5031, opened 2019-04-10
Missing index in IDN_OAUTH_CONSUMER_APPS table


mefarazath:
We need to add the below index on IDN_OAUTH_CONSUMER_APPS table

`CREATE INDEX IDX_IDN_OAUTH_CONSUMER_APPS ON IDN_OAUTH_CONSUMER_APPS(USERNAME,TENANT_ID,USER_DOMAIN, APP_NAME)`

When an OAuth application is created through DCR or through Management console, the OAuthApplicationMgtListener checks whether there is an already existing app with the same name at [1]. Without the index its a sequential search on the  IDN_OAUTH_CONSUMER_APPS. With the increase in the number of OAuth apps in IDN_OAUTH_CONSUMER_APPS table, this causes performance degradation and high CPU usage on the DB server. 

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/ca4690d734709b108ef2df33a69ae3c9242febc8/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/internal/OAuthApplicationMgtListener.java#L553

mefarazath:
These changed need to be added to the migration client as well..

malithie:
Created [1] to track the fix for migration client.
Closing this as this is fixed

Issue #4891, opened 2019-04-01
Missing indexes in the database


pamodaaw:
#### Description
Following  indexes improve the performance in WSO2 IS

- `CREATE INDEX IF NOT EXISTS IDX_SIA_AI ON SP_INBOUND_AUTH(APP_ID);`

There are no indexes which will be hit by the following queries.
- `SELECT PROPERTY_KEY, PROPERTY_VALUE FROM IDN_OIDC_PROPERTY WHERE TENANT_ID = ? AND CONSUMER_KEY = ?`
- `SELECT UM_ID FROM UM_HYBRID_ROLE WHERE UM_ROLE_NAME =? AND UM_TENANT_ID=?`

#### Type
Improvement

#### Affected version
5.7.0

mefarazath:
These changed need to be added to the migration client as well..

malithie:
Migration impact tracked with https://github.com/wso2/product-is/issues/5245

Resolving the issue as fixes are incorporated

PR #5135, opened 2019-04-25
Newly added Indexes were added to the IS migration client DBScrips

Buddhimah:
Newly added Indexes were added to the IS DBScrips
Related Issues:
wso2/product-is#4891
wso2/product-is#5031


modules/migration/migration-resources/5.8.0/dbscripts/step1/identity/db2.sql
```diff
@@ -24,3 +24,12 @@ CREATE INDEX IDX_USER_ID ON IDN_AUTH_USER_SESSION_MAPPING (USER_ID)
...
 /
+
+CREATE INDEX IDX_OCA_UM_TID_UD_APN ON IDN_OAUTH_CONSUMER_APPS(USERNAME,TENANT_ID,USER_DOMAIN, APP_NAME)
+/
+
+CREATE INDEX IDX_SPI_APP ON SP_INBOUND_AUTH(APP_ID)
+/
+
+CREATE INDEX IDX_IOP_TID_CK ON IDN_OIDC_PROPERTY(TENANT_ID,CONSUMER_KEY)
+/
```
tharindu-bandara: Shall we add a new line at the end of the file? Please do the same for others as well.

modules/migration/migration-resources/5.8.0/dbscripts/step1/identity/db2.sql
```diff
@@ -24,3 +24,12 @@ CREATE INDEX IDX_USER_ID ON IDN_AUTH_USER_SESSION_MAPPING (USER_ID)
 
 CREATE INDEX IDX_SESSION_ID ON IDN_AUTH_USER_SESSION_MAPPING (SESSION_ID)
 /
+
+CREATE INDEX IDX_OCA_UM_TID_UD_APN ON IDN_OAUTH_CONSUMER_APPS(USERNAME,TENANT_ID,USER_DOMAIN, APP_NAME)
```
mefarazath: We need to create the indexes if they don't already exist

mefarazath:
It would be better to check and only create indexes if they are not already present.
There may be cases where these indexes are already present.

modules/migration/migration-resources/5.8.0/dbscripts/step1/identity/db2.sql
Buddhimah: https://www.ibm.com/support/knowledgecenter/en/SSEPGG_11.1.0/com.ibm.db2.luw.sql.ref.doc/doc/r0000919.html

This is handling by default in db2 for other db types  changes Addressed

modules/migration/migration-resources/5.8.0/dbscripts/step1/identity/db2.sql
Buddhimah: done

PR #5331, opened 2019-05-24
User session management services through IS.

Chuhaa:
This PR contains changes in user dashboard to manage user session through IS.

Depends on:
Add user session api core component and stub #290


pamodaaw:
The session management feature will be added to the new user portal. The progress is tracked with the issue https://github.com/wso2/product-is/issues/6054.

Therefore this PR can be closed.

Issue #5277, opened 2019-05-13
Send authorization error to callback uri and not to oauth2_error.do page


IsuraD:
For some error reasons such as OAuthProblemException occurs, Identity Server redirect the error messages to the authenticationendpoint/oauth2_error.do URL. 

According to the OIDC specification [1], unless the redirection URI is invalid, the response should send the redirect URL.

`Unless the Redirection URI is invalid, the Authorization Server returns the Client to the Redirection URI specified in the Authorization Request with the appropriate error and state parameters. Other parameters SHOULD NOT be returned.`

An example scenario is provie a inalid response_type in authorize request.

PR #5366, opened 2019-06-04
Add tests to verify the OAuth2 error redirection

pamodaaw:
Add integration tests to validate the behavior of OAuth response redirection when an exception occurs with the default configurations.

- Tests the functionality implemented through https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1118
- Related to the issue https://github.com/wso2/product-is/issues/5277


mefarazath:
Can we link the git issue here please?

PR #6078, opened 2019-08-06
update identity inbound oauth version to fix test failures in 5.8.0

ShanikaWickramasinghe:
Update identity inbound oauth version to fix scenario test failures in IS 5.8.0

### DIFF
--- a/product-scenarios/pom.xml
+++ b/product-scenarios/pom.xml
@@ -98,7 +98,7 @@
         <version.commons.logging>1.1.1</version.commons.logging>
         <identity.inbound.auth.saml.version>5.4.31</identity.inbound.auth.saml.version>
         <identity.user.ws.version>5.1.11</identity.user.ws.version>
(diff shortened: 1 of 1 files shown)


product-scenarios/pom.xml
```diff
@@ -98,7 +98,7 @@
         <version.commons.logging>1.1.1</version.commons.logging>
         <identity.inbound.auth.saml.version>5.4.31</identity.inbound.auth.saml.version>
         <identity.user.ws.version>5.1.11</identity.user.ws.version>
-        <identity.inbound.auth.oauth.version>6.0.103</identity.inbound.auth.oauth.version>
+        <identity.inbound.auth.oauth.version>6.0.168</identity.inbound.auth.oauth.version>
```
ashensw: Why don't we update to the latest version?

ShanikaWickramasinghe: updated for 6.1.18

PR #6152, opened 2019-08-19
Update kernel to 4.5.0-alpha

senthalan:
This PR will update the kernel version to 4.5.0-alpha, which include log4j2 update, tomcat update and OSGi update.

In this PR the log4j2 from the kernel is replaced, in future we need to append our new log4j2 configs to the log4j2 from kernel without updating


modules/distribution/src/assembly/bin.xml
```diff
@@ -943,13 +958,13 @@
...
             <outputDirectory>wso2is-${pom.version}/repository/conf/identity</outputDirectory>
         </file>
-        <file>
-            <source>
-                ../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/repository/conf/etc/logging-config.xml
-            </source>
-            <outputDirectory>${pom.artifactId}-${pom.version}/repository/conf/etc/
-            </outputDirectory>
-        </file>
+        <!--<file>-->
```
madurangasiriwardena: Why commented out code?

GDRDABARERA:
Please upgrade the SAML version after with the release version after https://github.com/wso2-extensions/identity-inbound-auth-saml/pull/248 and merge this

senthalan:
Tests passed with https://jenkins-support-all.wso2.com/jenkins/job/DEV_PR_BUILDER/133/

PR #6362, opened 2019-09-05
Enable SAMLIdPInitiatedSLOTestCase

senthalan:
Previously this test was commented due to the logging UI removed. Now we use logging.view to get the logs


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/saml/SAMLIdPInitiatedSLOTestCase.java
```diff
@@ -98,20 +98,42 @@ public void testInit() throws Exception {
...
+        // Waiting to apply the log4j2 changes
+        Thread.sleep(10000);
+    }
+
+    private void resetISConfiguration() throws Exception {
+
+        log.info("Replacing log4j2.properties with default configurations");
+        serverConfigurationManager.restoreToLastConfiguration(false);
+        // Waiting to apply the log4j2 changes
+        Thread.sleep(10000);
```
ayshsandu: Why thread sleep? shall we periodically test whether changes are applied?

PR #6534, opened 2019-09-27
Update carbon.xml.j2 location

senthalan:


ayshsandu:
Same is fixed in https://github.com/wso2/product-is/pull/6535

Issue #6690, opened 2019-10-22
Toml configs for catalina-server.xml valve configs not properly templated. 


Buddhimah:
https://is.docs.wso2.com/en/5.9.0/administer/working-with-product-observability/

according to the above documentation, it should be capable of 
Change the RequestCorrelationIdValve in the Catalina-server.xml file 
from 

```
<Valve className="org.wso2.carbon.tomcat.ext.valves.RequestCorrelationIdValve"
    headerToCorrelationIdMapping="{'customHeader':'Correlation-ID'}" queryToCorrelationIdMapping="{'RelayState':'Correlation-ID'}"/>
```
to the value given below.

```
    <Valve className="org.wso2.carbon.tomcat.ext.valves.RequestCorrelationIdValve"
                            headerToCorrelationIdMapping="{'customHeader1':'Correlation-ID', 'customHeader2':'Second-Correlation-ID'}" queryToCorrelationIdMapping="{'RelayState':'Correlation-ID'}"/>
```

As this is not templated this capability is not supported

PR #6692, opened 2019-10-22
Toml configs for catalina-server.xml required valve configs templated

Buddhimah:
Fixed: https://github.com/wso2/product-is/issues/6690

RequestCorrelationIdValve brought to top at <Host> tag according to the documentation

https://is.docs.wso2.com/en/5.9.0/administer/working-with-product-observability/

### DIFF
--- a/modules/distribution/src/repository/resources/conf/key-mappings.json
+++ b/modules/distribution/src/repository/resources/conf/key-mappings.json
@@ -10,6 +10,7 @@
 
   "user_store.connection_url":"user_store.properties.ConnectionURL",
   "user_store.connection_name":"user_store.properties.ConnectionName",
-  "user_store.connection_password":"user_store.properties.ConnectionPassword"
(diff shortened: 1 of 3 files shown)


modules/distribution/src/repository/resources/conf/key-mappings.json
```diff
@@ -10,6 +10,7 @@
 
   "user_store.connection_url":"user_store.properties.ConnectionURL",
   "user_store.connection_name":"user_store.properties.ConnectionName",
-  "user_store.connection_password":"user_store.properties.ConnectionPassword"
+  "user_store.connection_password":"user_store.properties.ConnectionPassword",
 
+  "catalina.valves.multiple_header_config": "catalina_valves.request_correlation_id_valve.header_to_correlation_id_mapping"
```
hasinidilanka: We can directly give a meaningful name without giving a key-mapping.

Buddhimah: done

PR #6866, opened 2019-11-14
Fix adaptive templates metadata test failure

vihanga-liyanage:


madurangasiriwardena:
JsonNode.equals will treat two list elements are as equal only if they have the same values in the exact same order. So this will not fix the issue.

madurangasiriwardena:
Fixed with https://github.com/wso2/product-is/pull/6868

PR #6961, opened 2019-11-28
Remove dashboard from the product-is distribution.

thanujalk:


modules/integration/tests-integration/tests-backend/src/test/resources/testng.xml
```diff
@@ -24,7 +24,7 @@
             <class name="org.wso2.identity.integration.test.user.export.UserInfoExportTestCase" />
             <class name="org.wso2.identity.integration.test.consent.SelfSignUpConsentTest"/>
             <class name="org.wso2.identity.integration.test.oauth2.OIDCDiscoveryTestCase"/>
-            <class name="org.wso2.identity.integration.test.identity.governance.AdminForcedPasswordResetTestCase"/>
+            <!--<class name="org.wso2.identity.integration.test.identity.governance.AdminForcedPasswordResetTestCase"/>-->
```
darshanasbg: Have to fix test cases, rather commenting.

modules/integration/tests-integration/tests-backend/src/test/resources/testng.xml
```diff
@@ -70,7 +70,7 @@
             <class name="org.wso2.identity.integration.test.entitlement.EntitlementRestServiceTestCase"/>
             <class name="org.wso2.identity.integration.test.requestPathAuthenticator.RequestPathBasicAuthenticationSSOTest" />
             <class name="org.wso2.identity.integration.test.idp.mgt.IdentityProviderMgtServiceTestCase" />
-            <class name="org.wso2.identity.integration.test.dashboard.IDENTITY5811TestCase" />
+            <!--<class name="org.wso2.identity.integration.test.dashboard.IDENTITY5811TestCase" />-->
```
darshanasbg: Have to fix test cases, rather commenting.

modules/distribution/src/assembly/bin.xml
```diff
@@ -1007,16 +1007,6 @@
             </source>
             <outputDirectory>wso2is-${pom.version}/repository/conf</outputDirectory>
         </file>
-
-        <file>
-            <source>
-                ../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/repository/conf/identity/service-providers/sp_dashboard.xml
```
darshanasbg: We should remove the `sp_dashboard.xml` from the original feature as well.

modules/integration/tests-integration/tests-backend/src/test/resources/testng.xml
thanujalk: These test cases are testing the dashboard...

modules/distribution/src/assembly/bin.xml
thanujalk: The feature is removed so no longer we have sp_dashbard.xml in the pack.

Issue #7049, opened 2019-12-10
Enable User Application API integration tests


mefarazath:
Tests for User Application APIs have been added with https://github.com/wso2/product-is/pull/6839/files

But seems like they are not enabled in the testng.xml

PR #7099, opened 2019-12-15
Enable User Discoverable Application Test Cases

tharindu-bandara:
> Resolves https://github.com/wso2/product-is/issues/7049.

- Use dummy image and access urls for service provider creation tests.
- Enable tests in the testng.xml file.

### DIFF
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/user/application/v1/UserDiscoverableApplicationServiceTestBase.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/user/application/v1/UserDiscoverableApplicationServiceTestBase.java
@@ -44,6 +44,8 @@ public class UserDiscoverableApplicationServiceTestBase extends RESTAPIUserTestB
     protected static int TOTAL_DISCOVERABLE_APP_COUNT = 13;
     private static final String APP_NAME_PREFIX = "APP_";
     private static final String APP_DESC_PREFIX = "This is APP_";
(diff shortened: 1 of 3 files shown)


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/user/application/v1/UserDiscoverableApplicationServiceTestBase.java
```diff
@@ -95,6 +97,8 @@ private void createServiceProviders() throws Exception {
             serviceProvider = appMgtclient.getApplication(APP_NAME_PREFIX + i);
             if (serviceProvider != null) {
                 serviceProvider.setDiscoverable(true);
+                serviceProvider.setImageUrl(APP_IMAGE_URL);
```
tharindu-bandara: @malithie @darshanasbg 

These dummy URLs are used as the service providers are set as discoverable. I assumed that we are not testing the flows where image URLs and access URLs used.

malithie: We have to validate the set URLs in API response path as well

tharindu-bandara: Addressed in a52a880ae74fc40dd0480e1eb5f48c64eb1ea79c.

PR #7468, opened 2020-02-04
Bump auth-oauth version

tharindu-bandara:


tharindu-bandara:
Addressed with https://github.com/wso2/product-is/pull/7469.

PR #7678, opened 2020-02-17
Added integration test for ws-trust scenario.

deshankoswatte:
- Added a client which creates a RSTs with the help of a set of RST templates and calls the endpoint and receives the RSTR.
- Added integration test for scenarios: Issue, Renew and Validate a Security Token. These tests validates whether the received RSTRs are in valid format with the help of a set of RSTR templates.


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/sts/ActiveSTSTestCase.java
```diff
@@ -0,0 +1,236 @@
...
+        Assert.assertTrue(xPaths.containsAll(expectedValues),
+                "Expected changes did not match with the actual changes.");
+        Assert.assertEquals(xPaths.size(), expectedNoOfChanges,
+                "Expected " + expectedNoOfChanges + " changes "
+                        + "but found " + xPaths.size() + " changes.");
+    }
+
+    private void printQuery(String message, String query) {
+
+        System.out.println(message + ": \n" + query + "\n");
```
madurangasiriwardena: let's remove this.

modules/integration/tests-integration/tests-backend/src/test/resources/ws-trust-templates/request_security_token_RST.xml
```diff
@@ -0,0 +1,56 @@
+<soapenv:Envelope xmlns:soapenv="http://www.w3.org/2003/05/soap-envelope">
```
madurangasiriwardena: can we add license headers to these resource files?

deshankoswatte: These files are used for the test if we add the license header then the tests would fail.

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/sts/ActiveSTSTestCase.java
deshankoswatte: Remove unused method.

PR #7716, opened 2020-02-20
Test for /applications PATCH

mefarazath:
Adds tests for PATCH operations on an application using the REST API


thanujalk:
@mefarazath can we change this to PR and merge?

mefarazath:
We should be able to do it.

PR #8021, opened 2020-04-03
Upgrade carbon commons version

denuwanthi:


modules/p2-profile-gen/pom.xml
```diff
@@ -456,6 +456,11 @@
                                    org.wso2.carbon.identity.local.auth.fido:org.wso2.carbon.identity.application.authenticator.fido2.server.feature:${identity.local.auth.fido.version}
                                 </featureArtifactDef>
 
+                                 UUID Generator feature
```
ashensw, line 459: This needs to be commented?

denuwanthi: fixed with e9fc3560d1309c06353331a6b28f11f9c9b43bae

PR #8362, opened 2020-05-19
Introduce template properties to enable/disable http/https transport.

hasinthaindrajee:
Introduce template properties to enable/disable http/https transport.

### DIFF
--- a/modules/distribution/src/repository/resources/conf/default.json
+++ b/modules/distribution/src/repository/resources/conf/default.json
@@ -24,5 +24,8 @@
   "log_correlation.header_config": "{'activityid':'Correlation-ID'}",
 
   "monitoring.jmx.rmi_server_start": false,
-  "product.version" : "${product.version}"
+  "product.version" : "${product.version}",
+
+  "transport.http.enabled" : true,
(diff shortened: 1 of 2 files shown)


modules/distribution/src/repository/resources/conf/default.json
```diff
@@ -24,5 +24,8 @@
   "log_correlation.header_config": "{'activityid':'Correlation-ID'}",
 
   "monitoring.jmx.rmi_server_start": false,
-  "product.version" : "${product.version}"
+  "product.version" : "${product.version}",
+
+  "transport.http.enabled" : true,
+  "transport.https.enabled" : true
```
mefarazath, line 30: Do we need to allow disabling https?

hasinthaindrajee: There are customers who run http only mode who terminates https in LB level. They might need this config.

ruwanta: Can we make
 "transport.http.enabled" : false,

at defaults.json
Those who want to enable can do so. But always better to use https even for internal traffic.

Issue #7906, opened 2020-03-17
Enable the symmetric encryption configuration to be packed by default in WSO2 IS


denuwanthi:
**Description:**

The current product distribution has asymmetric encryption as the default.
This task is to enable the configurations related to making symmetric encryption as the default and pack it in the product distribution.

PR #8385, opened 2020-05-23
Make symmetric encryption as the default encryption mechanism forinte…

denuwanthi:
…rnal data

Resolves https://github.com/wso2/product-is/issues/7906


denuwanthi:
@ashensw @malithie Can we review and approve this?

modules/distribution/src/repository/resources/conf/default.json
```diff
@@ -27,5 +28,10 @@
   "product.version" : "${product.version}",
 
   "transport.http.enabled" : true,
-  "transport.https.enabled" : true
+  "transport.https.enabled" : true,
+
+  "system.parameter.\u0027org.wso2.CipherTransformation\u0027": "AES/GCM/NoPadding",
```
malithie, line 33: why we need to add a unicode here

denuwanthi: In toml keys are seperated by full stop (.)
Since org.wso2.CipherTransformation is one key, we need to define it as one key. We can use "system.parameter.'org.wso2.CipherTransformation'" or "system.parameter.\u0027org.wso2.CipherTransformation\u0027" . Since we have used unicode based one in other places [1], used the same.

[1]https://github.com/wso2/carbon-kernel/blob/e510cc9334f516d02b27aef3ddb670bf13ba45e9/distribution/kernel/carbon-home/repository/resources/conf/default.json#L174

Issue #8474, opened 2020-06-09
Typo in naming Apache


cpesch:
https://github.com/wso2/product-is/blob/45a0fb4ced8e968668120a02c8fb5a02b8876e59/README.txt#L214

should be Apac**h**e Rampart with an h instead of a g

dewniMW:
@cpesch Thanks for the reporting the typo.

PR #8475, opened 2020-06-09
Fix typo in README.txt

dewniMW:
### Issues
Fixes https://github.com/wso2/product-is/issues/8474


mefarazath:
The same typo exists on README.md fail as well. Shall we fix that as well?

cpesch:
Yes ;-)

Issue #9357, opened 2020-09-02
Assertion signing is not enabled by default when a creating a SAML application via the Application Mgt REST API


gayashanbc:
### Description 
Assertion signing is not enabled[1] by default when a creating a SAML application via the Application Mgt REST API.

[1] https://github.com/wso2/identity-api-server/blob/61c58f9c3daf7bc9ce32fe5c8c63fada52fda397/components/org.wso2.carbon.identity.api.server.application.management/org.wso2.carbon.identity.api.server.application.management.v1/src/main/java/org/wso2/carbon/identity/api/server/application/management/v1/core/functions/application/inbound/saml/ApiModelToSAMLSSOServiceProvider.java#L65

mefarazath:
There are two ways to fix this,

1. Set the relevant flag in the DTO as done in https://github.com/wso2/identity-api-server/pull/200
2. Fix the SAML inbound to not consider this propertly at all to decide whether to sign assertions or not. According to the spec, ```The <Assertion> element(s) in the <Response> MUST be signed``` therefore we should not reply on any property to decide this behaviour

mefarazath:
@thanujalk thoughts?

thanujalk:
@mefarazath yes, we should sign it by default.

PR #9993, opened 2020-10-21
Modify SPMetadataTestCase due to enabling assertion signing by default

sachiniWettasinghe:
Related to https://github.com/wso2/product-is/issues/9357

With https://github.com/wso2-extensions/identity-inbound-auth-saml/pull/310 the SAML assertions are set to be signed by default. Hence the test case is modified to reflect that behaviour.


pulasthi7 (approved):

sachiniWettasinghe:
SAML related integration tests are passing locally. Hence merging the PR.

Issue #10201, opened 2020-10-30
Warn logs during server start up


pulasthi7:
Following warn logs can be observed in the server startup with OOTB configurations (5.11.0-beta3)

```
WARN {javax.mail} - expected resource not found: /META-INF/javamail.default.address.map
```

```
WARN {org.apache.axis2.description.java2wsdl.DefaultSchemaGenerator} - We don't support method overloading. Ignoring [doSingleLogout]
```

```
WARN {org.wso2.carbon.identity.event.internal.IdentityEventServiceComponent} - Properties for userOperationDataDASPublisher is not configured. This event handler will not be activated
```

```
WARN {org.wso2.carbon.identity.event.internal.IdentityEventServiceComponent} - Properties for liteUserRegistration is not configured. This event handler will not be activated
```

piraveena:
`WARN {org.wso2.carbon.identity.event.internal.IdentityEventServiceComponent} - Properties for liteUserRegistration is not configured. This event handler will not be activated 
` is fixed with this PR https://github.com/wso2/carbon-identity-framework/pull/3177. This log is no longer available in IS5.11.0-beta6-snapshot pack

PR #10506, opened 2020-11-13
Add log level for axis2-description

piraveena:
Partial fix for https://github.com/wso2/product-is/issues/10201.

This PR is to  get rid of the below warn logs. 
`WARN {org.apache.axis2.description.java2wsdl.DefaultSchemaGenerator} - We don't support method overloading. Ignoring [doSingleLogout]`

Since there are two overloaded methods in SAMLSSOService component, the below warn logs are getting printed. Need to deprecate on method.   https://github.com/wso2-extensions/identity-inbound-auth-saml/pull/312

With this PR, a new log level is introduced. After we remove one of the overloaded method, we can revert this change back.


pulasthi7 (approved):

piraveena:
Builder is running here https://github.com/wso2/product-is/actions/runs/361384569

PR #10876, opened 2021-01-04
Add integration tests for token revocation when session is terminated via RESTapi

piraveena:
This needs to be merged after bumping the oauth version (6.4.122) in the product-is


piraveena:
Moved to https://github.com/wso2/product-is/pull/10888

Issue #10998, opened 2021-01-19
Move TOTP connector to plugins


mefarazath:
Part of https://github.com/wso2/product-is/issues/10612

As part of this effort we are,

- Moving the TOTP authenticator into plugins (Install as a feature)
- Install the org.wso2.carbon.extension.identity.helper via the feature (earlies this was copied to dropins)
- Move the TOTP pages to the authentication portal and stop packing the TOTP web apps

mefarazath:
There are few affected paths for a user migrating a user from 5.11.0 to 5.12.0

1. The connector is no longer added to the dropins folder
2. The totpauthethentication endpoint is no longer packed with the product. Users having customized totp pages have two choices.

* Merge the customziations to the authentication portal app. The pages related to TOTP are
- enableTOTP.jsp (TOTP enroll page)
- totp.jsp (TOTP login page)
- totpError.jsp (TOTP error page)

The 18n keys are moved to the i18n property file of the authentication portal

PR #11005, opened 2021-01-19
Move TOTP into plugins

mefarazath:
Part of #10998

- Remove totp connector from dropins
- Remove totp webapps from being added (TOTP page are now in authentication portal)
- Install TOTP via feature into plugins
- Install Identity Extension helper via feature into plugins

### DIFF
--- a/modules/p2-profile-gen/pom.xml
+++ b/modules/p2-profile-gen/pom.xml
@@ -249,6 +249,12 @@
                                 <featureArtifactDef>
                                     org.wso2.carbon.identity.carbon.auth.mutualssl:org.wso2.carbon.identity.authenticator.mutualssl.feature:${identity.carbon.auth.mutual.ssl.version}
                                 </featureArtifactDef>
+                                <featureArtifactDef>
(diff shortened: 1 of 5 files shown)


modules/p2-profile-gen/pom.xml
```diff
@@ -945,6 +951,17 @@
...
                                     <version>${org.wso2.carbon.identity.developer.version}</version>
                                 </feature>
+
+
+                                <feature>
+                                    <id>org.wso2.carbon.extension.identity.helper.feature.group</id>
+                                    <version>${identity.extension.utils}</version>
+                                </feature>
+                                <feature>
+                                    <id>org.wso2.carbon.extension.identity.authenticator.totp.connector.feature.group</id>
```
pulasthi7, line 961: let's remove the word 'connector' as it will no longer be a connector

pulasthi7 (approved):

modules/p2-profile-gen/pom.xml
mefarazath: Tracked this via https://github.com/wso2/product-is/issues/11006

Issue #11718, opened 2021-05-10
Remove always enabling Captcha filter and template it to enable when recaptcha is enabled


ashans:
**Describe the issue:**
Currently `CaptchaFilter` is always enabled in `/repository/conf/tomcat/carbon/WEB-INF/web.xml`. Need to remove that and add to `/repository/resources/conf/templates/repository/conf/tomcat/web.xml.j2` so that it will be enabled only when `recaptcha.enabled = true` is present as a configuration.

PR #11719, opened 2021-05-10
Add templated captcha filter and update governance and framework versions

ashans:
Add templated recaptcha filter to enable only when `recaptcha.enable` property is true.
Add recovery flow url pattern for Captcha Filter.
Updated identity-governance and identity framework versions

Resolves https://github.com/wso2/product-is/issues/11718


senthalan (approved):

senthalan:
Let's improve the documentation where this filter config is referenced. https://is.docs.wso2.com/en/latest/learn/configuring-recaptcha-for-single-sign-on/

ashans:
> Let's improve the documentation where this filter config is referenced. https://is.docs.wso2.com/en/latest/learn/configuring-recaptcha-for-single-sign-on/

Will be fixed in https://github.com/wso2/docs-is/pull/2592

PR #12084, opened 2021-07-09
Add dependency updater work flow

AnuradhaSK:
dependency-updater.yml - for update dependencies for product-is
fork-branch-deleter.yml - Delete the forked branches due to dependency updater when close/merge the PR

### DIFF
--- a/.github/workflows/dependency-updater.yml
+++ b/.github/workflows/dependency-updater.yml
@@ -1,0 +1,98 @@
+# This workflow will update dependencies for product-is.
+name: Dependency Updater
+on:
+  workflow_dispatch:
+  schedule:
+    # Everyday at 15:00 - 15:15 UTC (8.30-8.45 PM SL time)
+    - cron:  '0-15 15 * * *'
+
+env:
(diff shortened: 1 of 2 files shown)


senthalan (approved):

.github/workflows/dependency-updater.yml
```diff
@@ -0,0 +1,98 @@
...
+        with:
+          name: mvn-build.log
+          path: |
+            mvn-build.log
+          if-no-files-found: warn
+      - name: Clean untracked files
+        run: |
+          git clean -f -d
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
```
janakamarasena, line 84: Is it possible to use a cURL of other direct ways from git to create a PR. As this third-party github action will have access to the PAT. Lets avoid that if possible.

janakamarasena (approved):

.github/workflows/dependency-updater.yml
AnuradhaSK: addressed in 087cc73

PR #12488, opened 2021-09-10
Update opensaml version.

Yasara-Y:
Update opensaml version to 3.3.1.wso2v3


somindatommy:
@Yasara-Y this PR has been open for a long time. Can we get this resolved soon?

Yasara-Y:
This version bump is not required since this opensaml version in the pom file affects only during the compile time but not to the product and the opensaml jar in plugins of the product already gets updated with the effort to upgrade opensaml dependent feature.xml files after the release of new opensaml orbit version 3.3.1.wso2v3 .

Issue #10380, opened 2020-11-08
proxy_context_path does not work as expected


rksk:
**Describe the issue:**
The configuration "proxy_context_path" which is toml equivalent for "ProxyContextPath" in XML based configs; is not working as expected. The carbon management console does not work due to a hardcoded CSRF guard javascript path. All the react web apps are not working because they have not taken this config into account.

**Steps to reproduce**
IS configs
```
[server]
hostname = "idp.wso2.com"
node_ip = "127.0.0.1"
base_path = "https://$ref{server.hostname}"
proxy_context_path = "auth"

[transport.https.properties]
proxyPort = 443 
```

Nginx configs
```
server {
	listen 443;
	server_name idp.wso2.com;
	client_max_body_size 100M;

	root /usr/share/nginx/www;
	index index.html index.htm;

	ssl on;
	ssl_certificate /etc/nginx/ssl/server.crt;
	ssl_certificate_key /etc/nginx/ssl/server.key;

	location /auth/ {
		proxy_set_header X-Forwarded-Host $host;
		proxy_set_header X-Forwarded-Server $host;
		proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
		proxy_set_header Host $http_host;
		proxy_read_timeout 5m;
		proxy_send_timeout 5m;

		proxy_pass https://127.0.0.1:9443/;
		proxy_cookie_path / /auth;
		proxy_redirect https://idp.wso2.com/ https://idp.wso2.com/auth/;

		proxy_http_version 1.1;
		proxy_set_header Upgrade $http_upgrade;
		proxy_set_header Connection "upgrade";
	}
}
```
Generate certificates by following "Create SSL certificates" topic of https://docs.wso2.com/display/CLUSTER44x/Configuring+NGINX

Try logging into any of followings using the base URL as https://idp.wso2.com/auth/
- carbon
- user-portal (5.10.0)
- myaccount (5.11.0)
- console (5.11.0)

**Environment information**
 - Product Version:  IS 5.10.0, IS 5.11.0 beta4

**Related**
According to the mail "Deprecating WebContextRoot configuration from WSO2 Identity Server" in dev mailing list, we are going to depreciate WebContextRoot and continue with ProxyContextPath.

rksk:
Additionally, we need to document this once all fixed.

IakimLynnyk-TomTom:
Hi! Is there any chance for it to be implemented in products-is 5.0.11.0?

[...]

PR #12724, opened 2021-10-28
Include tomcat/carbon/WEB-INF/web.xml.j2 template in the product

chamathns:
$subject.
Include the `web.xml.j2` template in the `repository/resources/conf/templates/repository/conf/tomcat/carbon/WEB-INF` directory.
Related to https://github.com/wso2/product-is/issues/10380

### DIFF
--- a/modules/distribution/pom.xml
+++ b/modules/distribution/pom.xml
@@ -417,9 +417,6 @@
 
                                 <replace file="../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/repository/conf/registry.xml" token="&lt;versioningRatings&gt;true&lt;/versioningRatings&gt;" value="&lt;versioningRatings&gt;false&lt;/versioningRatings&gt;" />
 
-                                <delete file="../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/repository/resources/conf/templates/repository/conf/tomcat/carbon/WEB-INF/web.xml.j2" />
-                                <delete dir="../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/repository/resources/conf/templates/repository/conf/tomcat/carbon/WEB-INF/" />
-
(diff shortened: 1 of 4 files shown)


IsuraD:
> @pulasthi7 @IsuraD @senthalan The default values for the configurable properties have been added to the `modules/distribution/src/repository/resources/conf/default.json` in this PR. What should be the ideal place these default values should be put in?

I think it is fine to add default.json

IsuraD (approved):

pulasthi7:
Since this is coming from kernel, the file should be templated at the kernal. Is there any particular reason to do it in the product?

We might miss the changes introduced at the kernel level if we go this path

chamathns:
> Since this is coming from kernel, the file should be templated at the kernal. Is there any particular reason to do it in the product?
> 
> We might miss the changes introduced at the kernel level if we go this path

The reason for not templating it at the kernel level was to make sure the tomcat/carbon/WEB-INF/web.xml file in 

> Since this is coming from kernel, the file should be templated at the kernal. Is there any particular reason to do it in the product?
> 
> We might miss the changes introduced at the kernel level if we go this path

The reason why I went with product-level templating was to have the flexibility of adding product-specific configs and default values for the templated file. But, yes, I agree, by going down this path, we would miss changes introduced to this file at the kernel level.

chamathns:
> Since this is coming from kernel, the file should be templated at the kernal. Is there any particular reason to do it in the product?
> 
> We might miss the changes introduced at the kernel level if we go this path

The reason why I went with product-level templating was to have the flexibility of adding product-specific configs and default values for the templated file. But, yes, I agree, by going down this path, we would miss changes introduced to this file at the kernel level.

chamathns:
Closing this PR since the changes with https://github.com/wso2/carbon-kernel/pull/3115 would do the templating at the kernel level as suggested by @pulasthi7

Issue #12853, opened 2021-12-08
Remove artifacts packed inside features folder in wso2is distribution. 


senthalan:
**Describe the issue:**
When building the wso2is distribution in IS, we are coping some files(jars, config files, config template files) from the features to the respective folders (lib/runtimes, conf, resources). But even after the coping we are not clearing these files and get packed with the distribution. 

This leads to the following issues
- Security scans identify them as vulnerable jars(even though they will be not loaded to the runtime), so we need to provide fixes for these as well.
-  The distribution pack size. (features folder takes up to ~29 MB)

If we can clear these artifacts once they are copied to the needed location, we can solve these.

PR #12861, opened 2021-12-08
Remove artifacts from distribution feature folder

senthalan:
Fix https://github.com/wso2/product-is/issues/12853

### DIFF
--- a/modules/distribution/src/assembly/dist.xml
+++ b/modules/distribution/src/assembly/dist.xml
@@ -35,7 +35,8 @@
                 <exclude>**/*.lock</exclude>
                 <exclude>**/.data</exclude>
                 <exclude>**/.settings</exclude>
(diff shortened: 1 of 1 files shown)


modules/distribution/src/assembly/dist.xml
```diff
@@ -35,11 +35,12 @@
                 <exclude>**/*.lock</exclude>
                 <exclude>**/.data</exclude>
                 <exclude>**/.settings</exclude>
-                <!--removing jars and wars from repository/components/feature folder-->
+                <!--removing jars,json, j2 and wars from repository/components/feature folder-->
```
omindu, line 38:
```suggestion
                <!--removing jars, json, j2 and wars from repository/components/feature folder-->

```

omindu (approved):

PR #13039, opened 2022-01-26
Add changes required with the keypair renewal

chamathns:
$subject


chamathns:
locally tested and verified the Keystore cert renewal with updated samples-is (v4.3.8) and kernel 4.7.0-alpha10.

ashendes (approved):

PR #13164, opened 2022-02-25
Bump kernel version to 4.7.0-m11

chamathns:
$subject


chamathns:
Included with https://github.com/wso2/product-is/pull/13140

PR #13622, opened 2022-05-05
Bump basic authenticator version

piyarathnalakmali:
$Subject


somindatommy (approved):

piyarathnalakmali:
Version is already bumped by the dependency updater.

Issue #13739, opened 2022-05-25
Apply XACML policies were failed 


aaujayasena:
**Describe the issue:**
Following behaviors can be observed in xacml policy scenarios.

**Issue 01:**
Given policy templated are not working in tryout but when create new policy and it work on tryout.

Try out with given policy
<img width="800" alt="Screenshot 2022-05-25 at 07 26 15" src="https://user-images.githubusercontent.com/39077751/170162970-6ad9b85c-b300-42a5-8b8b-2e2f0d8830e5.png">

https://user-images.githubusercontent.com/39077751/170163076-d177d806-9ce0-4456-961a-aaa6da45b725.mov

Tryout with new Policy created according to the document https://is.docs.wso2.com/en/latest/learn/fine-grained-authorization-with-xacml/
<img width="728" alt="image" src="https://user-images.githubusercontent.com/39077751/170164116-8cda3c64-1749-4e5b-9990-677b0b5e5bb3.png">
 

**Issue 02:**
Apply policy for application, failed the authentication. 

<img width="818" alt="image" src="https://user-images.githubusercontent.com/39077751/170173978-fafdda2e-21a0-4789-9035-7a8cd45b68dc.png">

**How to reproduce:**
Issue 01: 
1. Login to the management console
2. Navigate to policy administration and select the tryout in authn_time_based_policy_template
3. Add Resource as SP_NAME, Subject name as 10.00.00 and Action name as authenticate
4. Click test evaluate

Issue 02: 
1. Login to the management console
2. Navigate to policy administration and create policy according to the steps in document https://is.docs.wso2.com/en/latest/learn/fine-grained-authorization-with-xacml/
3.  select the tryout in created policy and tryout
4. Publish the policy
5. Create users and SP
6. Login to application with different permitted users

**Expected behavior:**
Issue 01:
According to the policy application authentication allows to time between 9.00.00-17.00.00. Hence 10.00.00 it should permit the authetication. 

Issue 02:
Permitted users should be authenticate and others should be deny. 

**Environment information** (_Please complete the following information; remove  any unnecessary fields_) **:**
 - Product Version: [e.g., IS 5.10.0, IS 5.9.0]
 - OS: [e.g., Windows, Linux, Mac]
 - Database: [e.g., MySQL, H2]
 - Userstore: [e.g., LDAP, JDBC]

---

### Optional Fields

**Related issues:**

**Suggested labels:**

piraveena:
Reason for the issue2:

[...]

PR #13745, opened 2022-05-25
Change role claim to roles in XACML policies and add a group based authentication template

piraveena:
Part of https://github.com/wso2/product-is/issues/13739

Changes in the PR:
- Change the `http://wso2.org/claims/role` to `http://wso2.org/claims/roles`
- Introduced a group based authentication XACML policy template


modules/distribution/conf/policies/authn_time_and_role_based_policy_template.xml
```diff
@@ -28,11 +28,11 @@
                 <Apply FunctionId="urn:oasis:names:tc:xacml:1.0:function:or">
                     <Apply FunctionId="urn:oasis:names:tc:xacml:1.0:function:string-is-in">
                         <AttributeValue DataType="http://www.w3.org/2001/XMLSchema#string">ROLE_1</AttributeValue>
```
ashensw, line 30: Shall we append the `Internal` domain to these scripts as well?

ashensw (approved):

PR #13895, opened 2022-06-09
Bump identity-apps version

piyarathnalakmali:
$subject


piyarathnalakmali:
Automatically upgraded with dependency updater.

PR #14126, opened 2022-07-06
Pr builder fix for jdk11

nirmal101:
Fixed the issue where PR-builder version is failing for different JDK versions.

### DIFF
--- a/.github/scripts/pr-builder.sh
+++ b/.github/scripts/pr-builder.sh
@@ -21,8 +21,8 @@ echo "::set-output name=REPO_NAME::$REPO"
 echo "=========================================================="
 echo "Cloning product-is"
 echo "=========================================================="
-echo " java home"
-echo $1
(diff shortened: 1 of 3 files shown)


piraveena (approved):

.github/scripts/pr-builder.sh
```diff
@@ -135,11 +143,18 @@ else
...
   echo "=========================================================="
+
+  if [ "$JDK_VERSION" = "11" ]; then
+    export JAVA_HOME=$JAVA_11_HOME
+  else
+    export JAVA_HOME=$JAVA_8_HOME
+  fi
+
   if [ "$REPO" = "carbon-kernel" ]; then
     mvn clean install -Dmaven.test.skip=true --batch-mode | tee mvn-build.log
```
piraveena, line 154: Do we need to skip testcases for kernel?

PR #14158, opened 2022-07-08
Bump identity apps version to v1.2.1030

JayaShakthi97:
> Bump identity apps version to v1.2.1030


JayaShakthi97:
Version is already bumped.

PR #14199, opened 2022-07-12
Update pr builder

nirmal101:
Old pr-builder that only supports jdk 8 has been removed.


Thumimku:
Get a pull from upstream and squash your commits

Thumimku (approved):
LGTM:  fix above comment

PR #14450, opened 2022-07-29
Update jks to support ScenarioTests

VivekVinushanth:
This PR updates the cert to support Scenario Testing.


piraveena:
what jks file is this? is this our default file? if not, tests may fail with local builds

piraveena (approved):

PR #14883, opened 2022-09-15
Remove workflow from IS pack

hwupathum:
Remove workflows from IS pack to make it a connector


hwupathum:
This if for IPK. Therefore need to be merged to a new branch

PR #15363, opened 2023-01-17
Remove workflow feature

hwupathum:
This PR contains the minimal changes required to remove workflow feature and its integration tests from product-is pack

PR #15480, opened 2023-02-08
Remove ciphertool and IS & worker profiles

hwupathum:
$subject


modules/p2-profile-gen/pom.xml
```diff
@@ -961,854 +957,6 @@
                             <artifactRepository>file:${basedir}/target/p2-repo</artifactRepository>
                         </configuration>
                     </execution>
-
-
-                    <!-- Only IS Profile (IS Front-End + IS Back-End) -->
-                    <execution>
-                        <id>creating-is-profile</id>
```
madurangasiriwardena, line 968: AFAIR, IS profile was not having the embedded workflow engine. That was the only change in IS profile compared to default profile. Have we already removed the embedded workflow engine from the default profile?

hwupathum: Yes it was removed in https://github.com/wso2/product-is/pull/15363

PR #15609, opened 2023-03-13
Add fossa scanner

RushanNanayakkara:
## Purpose
Adds the workflow for fossa scanner. The scan is to be run 12AM on each day.

### DIFF
--- a/.github/workflows/fossa-scanner.yaml
+++ b/.github/workflows/fossa-scanner.yaml
@@ -1,0 +1,14 @@
+name: "FOSSA Scanner"
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


.github/workflows/fossa-scanner.yaml
```diff
@@ -0,0 +1,14 @@
+name: "FOSSA Scanner"
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
```
RushanNanayakkara, line 6: workflow_dispatch is added for testing purposes. This can be removed later after flow is confirmed.

DMHP (approved):

Issue #15573, opened 2023-03-01
Update the signup API to support password less registration option


pamodaaw:
The current signup API only supports username and password registration. With the improvement suggested in the issue, https://github.com/wso2/product-is/issues/15449, the signup API can also support different credential options when user registration.

- [ ] BE improvements ( tracked with 15449)
- [ ] API layer changes
- [ ] Unit tests
- [ ] Integration tests
- [ ] Documentation update
- [ ] Update the UIs to support the registration

pamodaaw:
A new set of APIs will be introduced to support the signup flow configuration and perform the signup against the configured flow. This requirement will be tracked in the https://github.com/wso2/product-is/issues/16371. Hence closing this issue.

PR #15625, opened 2023-03-16
Update the governance version

pamodaaw:
Update the governance version to revert the changes done via https://github.com/wso2-extensions/identity-governance/pull/669

Related issue: https://github.com/wso2/product-is/issues/15573


pamodaaw:
The version is updated by the dependency updater PR

PR #15872, opened 2023-05-15
Update self service scope returing for client credential grant

nilasini:


nilasini:
The failure is an intermittent one. Hence merging the PR

Screenshot 2023-05-16 at 15 06 43

piraveena (approved):

PR #16159, opened 2023-06-29
Increasing the menu panel width to accommodate IdV Provider text.

nipunsampath:
### Purpose
$subject

### Related PRs

- https://github.com/wso2-extensions/identity-verification/pull/9

### DIFF
--- a/modules/styles/product/src/main/resources/web/styles/css/main.css
+++ b/modules/styles/product/src/main/resources/web/styles/css/main.css
@@ -12,6 +12,7 @@ table#main-table td#menu-panel {
 	border-right: solid 1px #989898;
 	background-position: right top;
 	background-repeat: repeat-y;
+	width: 230px;
 }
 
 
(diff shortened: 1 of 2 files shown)


modules/styles/product/src/main/resources/web/styles/css/main.css
```diff
@@ -12,6 +12,7 @@ table#main-table td#menu-panel {
 	border-right: solid 1px #989898;
 	background-position: right top;
 	background-repeat: repeat-y;
+	width: 230px;
```
brionmario, line 15: This is a global change that affects other places right?
IMO, this should be `min-width`.

nipunsampath: Yes, It affects the whole menu panel in IS Management console. Fixed with 2b9bcbd09c323e9049e1cf026cc132137dc3b907.

brionmario (approved):

PR #16233, opened 2023-07-12
Change default cookie specs provider to RFC6265CookieSpecProvider

shanggeeth:
$Subject

### Purpose

- Httpclient used in Integration tests was not using RFC6265CookieSpecProvider as default cookie specs provider.
In order to upgrade tomcat-catalina > 9.0.68, we need to set `RFC6265CookieSpecProvider` as default cookie specs provider for httpclient used in intgration tests.
- Change deprecated `DefaultHttpClient` usage to `ClosableHttpClient`
- Setting `RFC6265CookieSpecProvider` as default cookie specs provider handles cookies from Tomcat's Legacy Cookie Processor as well, hence this change is backward compatible
- The tests have passed locally with and without tomcat-catalina changes.


Thumimku (approved):
LGTM

PR #16297, opened 2023-07-21
Set system property jdk.util.zip.disableZip64ExtraFieldValidation to true

chamathns:
$subject

### DIFF
--- a/.github/scripts/pr-builder.sh
+++ b/.github/scripts/pr-builder.sh
@@ -59,7 +59,7 @@ if [ "$REPO" = "product-is" ]; then
 
   cat pom.xml
   export JAVA_HOME=$JAVA_11_HOME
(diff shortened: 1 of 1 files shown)


.github/scripts/pr-builder.sh
```diff
@@ -59,7 +59,7 @@ if [ "$REPO" = "product-is" ]; then
 
   cat pom.xml
   export JAVA_HOME=$JAVA_11_HOME
-  mvn clean install --batch-mode | tee mvn-build.log
+  mvn clean install -Dmaven.test.skip=true -Djdk.util.zip.disableZip64ExtraFieldValidation=true --batch-mode | tee mvn-build.log
```
shanggeeth, line 62: Do we need to skip tests here ?

chamathns: that's been added by mistake, this is fixed now

shanggeeth (dismissed):

shanggeeth (approved):

Issue #16158, opened 2023-06-28
Adaptive authentication adaptive.sh file execution issue.


melanisilva:
**Describe the issue:**
When enabling the Adaptive authentication feature, the adaptive.sh file doesn't stop execution when an erroneous issue has occurred. 

<img width="1077" alt="Screenshot 2023-06-27 at 13 30 43" src="https://github.com/wso2/product-is/assets/41188245/112df70d-007a-4b32-866b-827daaf6fc85">

**How to reproduce:**

**Expected behavior:**
The execution of the adaptive.sh file should be terminated when an error occurs.

**Environment information** (_Please complete the following information; remove  any unnecessary fields_) **:**
- Product Version: [e.g., IS 5.10.0, IS 5.9.0] - IS 6.2.0-alpha
- OS: [e.g., Windows, Linux, Mac] - Mac
- Database: [e.g., MySQL, H2] - H2 / MSSQL
- Userstore: [e.g., LDAP, JDBC]
- 

---

PR #16377, opened 2023-08-03
Fix adaptive authentication script file execution issue

Shenali-SJ:
### Purpose
This resolves issues encountered when downloads fail during the execution of adaptive.sh and adaptive.bat files.

### Related Issues
https://github.com/wso2/product-is/issues/16158

### DIFF
--- a/modules/distribution/src/bin/adaptive.bat
+++ b/modules/distribution/src/bin/adaptive.bat
@@ -132,10 +132,9 @@ if exist "%CARBON_HOME%\repository\components\lib\nashorn-core-*.jar" (
     echo Downloading required Nashorn library : nashorn-core-%NASHORN_VERSION%
 	  curl https://repo1.maven.org/maven2/org/openjdk/nashorn/nashorn-core/%NASHORN_VERSION%/nashorn-core-%NASHORN_VERSION%.jar -o %CARBON_HOME%/repository/components/lib/nashorn-core-%NASHORN_VERSION%.jar
     rem if download failed
-    set exit_code_n=%ERRORLEVEL%
-    if %exit_code_n% neq "0" (
-      echo Nashorn library update failed with exit code: %exit_code_n%
-      exit /b %exit_code_n%
+    if !ERRORLEVEL! neq 0 (
+      echo Nashorn library update failed with exit code: !ERRORLEVEL!
+      exit /b !ERRORLEVEL!
     )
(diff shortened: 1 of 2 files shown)


modules/distribution/src/bin/adaptive.sh
```diff
@@ -127,7 +127,11 @@ else
   else
     sever_restart_required=true
     echo "Nashorn library not found. Starting to download....."
-    curl https://repo1.maven.org/maven2/org/openjdk/nashorn/nashorn-core/$NASHORN_VERSION/nashorn-core-$NASHORN_VERSION.jar -o $CARBON_HOME/repository/components/lib/nashorn-core-$NASHORN_VERSION.jar
+    res_code=$(curl https://repo1.maven.org/maven2/org/openjdk/nashorn/nashorn-core/$NASHORN_VERSION/nashorn-core-$NASHORN_VERSION.jar -o $CARBON_HOME/repository/components/lib/nashorn-core-$NASHORN_VERSION.jar)
+    if test "$exit_code" != "0"; then
+      echo "Nashorn library download failed with exit code: $exit_code"
+      exit 1
+    fi
```
Kanapriya, line 134: Don't we need to add this changes to the adaptive.bat file?

Shenali-SJ: The related issue was reported for only the shell script. Shall we look into bat files as well?

melanisilva:
Verified the change in Windows OS 
Both scenarios are working as expected, Where the adaptive authentication scripts fail and pass.
success

modules/distribution/src/bin/adaptive.bat
```diff
@@ -155,12 +165,23 @@ if exist "%CARBON_HOME%\repository\components\lib\asm-util-*.jar" (
     call :removeLibrary "ASM Util", "dropins", "%CARBON_HOME%\repository\components\dropins\asm_util_!LOCAL_ASM_VERSION!*.jar" 
     echo Downloading required ASM-Util library : asm-util-%ASM_VERSION%
 	  curl https://repo1.maven.org/maven2/org/ow2/asm/asm-util/%ASM_VERSION%/asm-util-%ASM_VERSION%.jar -o %CARBON_HOME%/repository/components/lib/asm-util-%ASM_VERSION%.jar
+	rem if download failed
```
ayeshajay, line 168: Check indentation

PR #16540, opened 2023-08-28
Refactor: Switching to Service Account JSON Credentials

JayanaGunaweera01:
**Description:**

This pull request introduces a pivotal enhancement to the GitHub environment's mechanism for downloading GA packs from Google Drive during migration automation

**Changes Made:**

1.Transition to Service Account JSON Credentials:

The method for accessing Google Drive has been revamped. Instead of using a GCP API key, we now leverage a Service Account to download files. This is a more secure approach to handle file downloads from Google Drive. For this purpose, a dedicated file (creds.json) has been created temporarily to securely store the Service Account JSON credentials. creds.json  is deleted immediately after its intended use. This ensures that the credentials cannot be accessed post-operation, safeguarding against potential vulnerabilities.

2.Removal of Old GCP Credentials:

The older method of using the GCP API key, represented by the secrets GCP_CLIENT_ID, GCP_CLIENT_SECRET, and GCP_REFRESH_TOKEN, has been deprecated and removed from the workflow. This change aligns with the shift towards utilizing the Service Account for the file download operations.

### DIFF
--- a/.github/workflows/migration-automation.yml
+++ b/.github/workflows/migration-automation.yml
@@ -201,7 +201,7 @@ jobs:
         run: |
           chmod +x ${{ github.workspace }}/.github/migration-tester/migration-automation/ubuntu-os/migration-script-ubuntu.sh 
           sh ${{ github.workspace }}/.github/migration-tester/migration-automation/ubuntu-os/migration-script-ubuntu.sh "${{ github.event.inputs.urlOld }}" "${{ github.event.inputs.urlNew }}" "${{ github.event.inputs.currentVersion }}" "${{ github.event.inputs.migratingVersion }}" "${{ github.event.inputs.database }}" "${{ github.event.inputs.os }}" "${{ secrets.MIGRATION_EMAIL }}" "${{ secrets.MIGRATION_PASSWORD }}" "${{ secrets.MIGRATION_PAT }}" creds.json | tee "${{ github.workspace }}/.github/migration-tester/migration-automation/logs.txt"
-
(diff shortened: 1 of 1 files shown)


.github/workflows/migration-automation.yml
```diff
@@ -54,6 +54,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+      - name: Save service account JSON credentials to a file
+        run: echo '${{ secrets.MIGRATION_JSON }}' > ${{ github.workspace }}/.github/migration-tester/migration-automation/creds.json
```
chamathns, line 58: can writing the credentials to a file like this pose any security issues?

JayanaGunaweera01: @chamathns ,

Thank you for raising this valid concern. In the context of GitHub Actions workflows, files created within the runner environment – be it a temporary VM or container – are ephemeral. Once the workflow concludes, these files are purged, especially since our current procedure does not involve generating artifacts or saving logs from this step.

**Having said that, taking into account best practices and suggestions such as yours, I have transitioned our methodology from file-based storage to leveraging environment variables for credential management.** This strategic shift not only streamlines the process but also amplifies security, effectively circumventing the risks associated with file-based credential storage.

For a detailed overview of the changes implemented, I would direct your attention to commits 1b056c8 and 3db1330.

RushanNanayakkara:
Closing this PR since it's been inactive for a long time. Please reopen if needed.

Issue #16413, opened 2023-08-08
Password and code grant does not work for non unique id user stores after update


deshankoswatte:
**Describe the issue:**

Both the password and code grant does not work after updating the server to update level 256 in the WSO2 Identity Server 5.11.0.

**How to reproduce:**

- Configure a non unique id secondary user store with the WSO2 Identity Server.
- Try password grant or code grant with a user that exists in the configured user store.

An error will be thrown and the authentication fails.

_Password grant:_

```
[2023-08-08 16:26:03,723] [fb9d3131-2f9b-41ed-b6fc-0d7085b6f2a6] DEBUG {org.wso2.carbon.identity.oauth2.token.AccessTokenIssuer} - Error occurred while validating grant org.wso2.carbon.identity.oauth2.IdentityOAuth2Exception: Authentication failed for user1
	at org.wso2.carbon.identity.oauth2.token.handlers.grant.PasswordGrantHandler.validateUserCredentials(PasswordGrantHandler.java:210)
	at org.wso2.carbon.identity.oauth2.token.handlers.grant.PasswordGrantHandler.validateGrant(PasswordGrantHandler.java:103)
	at org.wso2.carbon.identity.oauth2.token.AccessTokenIssuer.issue(AccessTokenIssuer.java:245)
	at org.wso2.carbon.identity.oauth2.OAuth2Service.issueAccessToken(OAuth2Service.java:257)
	at org.wso2.carbon.identity.oauth.endpoint.token.OAuth2TokenEndpoint.issueAccessToken(OAuth2TokenEndpoint.java:308)
	at org.wso2.carbon.identity.oauth.endpoint.token.OAuth2TokenEndpoint.issueAccessToken(OAuth2TokenEndpoint.java:128)
	at org.wso2.carbon.identity.oauth.endpoint.token.OAuth2TokenEndpoint.issueAccessToken(OAuth2TokenEndpoint.java:109)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.cxf.service.invoker.AbstractInvoker.performInvocation(AbstractInvoker.java:179)
[...]

PR #16558, opened 2023-08-31
Add integration test for non-uniqueID userstores

hwupathum:
Add an integration tests for non-unique ID JDBC userstores

Related to https://github.com/wso2/product-is/issues/16413


Kanapriya (approved):

Kanapriya:
@hwupathum : Please check whether we can use the h2 scripts from the build distribution?. Because whenever have changes in h2 script, we may need to update the integration test too

hwupathum:
Non-uniqueID userstores were deprecated in IS 5.9 but still has the support in master. The db schema used by this test is different from the latest version, hence we cannot use the script from the build distribution. When the support for non-unique ID userstores are removed, we can remove this integration test

PR #16609, opened 2023-09-08
Set subject attribute in EmailOTPTestCase

hwupathum:
Enable subject identifier in first step of 2FA from EmailOTPTestCase


nilasini:
can you please attach the actual fix we introduced for this change?

hwupathum:
This integration test change is related to https://github.com/wso2-extensions/identity-outbound-auth-email-otp/pull/164

nilasini:
Can you please give some more details on why we need to make the change in integration test?

hwupathum:
In the PR https://github.com/wso2-extensions/identity-outbound-auth-email-otp/pull/164 I added a null check for authenticatedUser vaiable and while debugging found out that the authenticatedUser becomes null in the  EmailOTPTestCase, which is not the correct behaviour

nilasini:
Did we include this information(Attribute step should be enabled)  in our email OTP as first-factor documentation?

nilasini (approved):

hwupathum:
The emailOTP authenticator can be used as the sceond factor without enabling attribute step, hence the no need to change integration test

PR #16843, opened 2023-10-03
Bump identity-apps component versions

JayaShakthi97:
$subject


JayaShakthi97:
Versions are already updated.

PR #16995, opened 2023-10-17
Fix failing integration tests when tenant qualified urls enabled

Thisara-Welmilla:
Issue:
- https://github.com/wso2/product-is/issues/16999

Fixed failing integration test when tenant qualified urls are enabled.


chamathns (approved):
LGTM

PR #17220, opened 2023-10-26
Enable tenant qualified URLs feature by default

Thisara-Welmilla:
Enable tenant qualified URLs feature via default.json


chamathns:
there are merge conflicts. Let's fix that. otherwise LGTM

chamathns (approved):

PR #18094, opened 2023-11-27
Update pom.xml

PasinduYeshan:


AnuradhaSK:
bumped the version to higher version.hence closing

PR #18117, opened 2023-11-28
Integration tests for the fix for issue in claim mapping for token exchange and jwt grant flow

Lakshan-Banneheke:
### Related PR

- https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2270

The above PR fixes a bug in the claim mapping of the token exchange flow and the jwt grant flow. Previously claim mapping was not happening for SP requested claims if the IDP mapping was not present. This was fixed and this PR is to update the integration tests to reflect the fix.


Yasasr1 (approved):

Lakshan-Banneheke:
Tests will momentarily fail as these tests are dependant on https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2270

PR #18394, opened 2023-12-04
Modify extractClaims to work with regex

sahandilshan:
### Purpose
- $subject
- Modify the `extractClaims` method to select the div tags that contain the `claim-list` class. With the earlier implementation, the tests will fail if someone adds a newer class to the div tag. This implementation will check a regex match instead of a string comparison to identify the claim list.


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/util/Utils.java
```diff
@@ -357,38 +359,55 @@ public static HttpResponse sendPOSTConsentMessage(HttpResponse response, String
         post.setEntity(new UrlEncodedFormEntity(urlParameters));
         return httpClient.execute(post);
     }
-
+    
     private static List<String> extractClaims(HttpResponse response) throws IOException {
-
+
```
SujanSanjula96, line 364: shall we remove trailing spaces

SujanSanjula96 (approved):

PR #18590, opened 2023-12-14
Bump console version to 2.10.26

nipunsampath:
### Purpose 
$subject


nipunsampath:
Closing because the version is already bumped by the dependency updater.

PR #18770, opened 2024-01-02
[FAPI Conformance] Fix FAPI action failure due to consent flow skip

anjuchamantha:
FAPI conformance suite git action build was failing due to skipping of the consent flow.
Earlier when a new scope (for example abc) provided in /authorize request, user is prompted to a consent flow to allow or deny that. However with the new IS improvements, consent page is shown only for registered scope claims. Due to this FAPI action build was failing, and this PR fixes it.


Thumimku (approved):
LGTM

PR #18779, opened 2024-01-02
Add tomcat apr openssl config

hwupathum:
Use the following config to use APR implementation with OpenSSL for tomcat server.

```
[transport.https]
openssl_apr.enabled = true
```

**Note**
You need to install following libraries before enabling the config
- APR library
- JNI wrappers for APR used by Tomcat (libtcnative)
- OpenSSL libraries


hwupathum:
Duplicate of https://github.com/wso2/product-is/pull/20223

PR #19905, opened 2024-03-01
Improve device flow negative test cases

Thumimku:
This PR add some negative test cases when client id is invalid and not present in the request.


darshanasbg:
@Thumimku: Are you working on this? Can we get this change finalized?

Thumimku:
@darshanasbg : Yes I am working on this but not actively. This PR has some integration test failures. I will fix and rasie another one.

darshanasbg:
Ack.. Thanks..

PR #19988, opened 2024-03-12
Bump framework version

ZiyamSanthosh:
$subject


Avarjana:
We can skip the PR builder since this is only a addition of a constant

Avarjana (approved):

PR #20058, opened 2024-03-20
Bump kernel version to 4.10.11

sandushi:
### Purpose
$subject


darshanasbg:
Hi @sandushi, if you are to update the kernel, you need to `carbon.product` file as well.. You can refer a similar change in https://github.com/wso2/product-is/pull/19945.

Please update the PR with the necessary changes, and if you don't need to this change anymore, then lets close the PR.

Thanks,

PR #20300, opened 2024-04-30
Bump framework and identity inbound versions

RivinduM:
This PR bumps dependencies for identity-framework and identity-inbound-oauth


darshanasbg:
Already updated with https://github.com/wso2/product-is/pull/20302

PR #20353, opened 2024-05-10
Update error messages

hwupathum:
Update error messages of openssl-tls script

### DIFF
--- a/modules/distribution/src/bin/openssl-tls.sh
+++ b/modules/distribution/src/bin/openssl-tls.sh
@@ -47,7 +47,7 @@ configure_openssl_conf() {
 check_dependencies() {
     for cmd in "$@"; do
         if ! command -v "$cmd" >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


modules/distribution/src/bin/openssl-tls.sh
```diff
@@ -176,9 +176,15 @@ PRGDIR=$(dirname "$PRG")
 # Only set CARBON_HOME if not already set
 [ -z "$CARBON_HOME" ] && CARBON_HOME=$(cd "$PRGDIR/.." ; pwd)
 
+# Check if $CARBON_HOME has spaces
+if echo "$CARBON_HOME" | grep -q ' '; then
+    echo "[Error]: Please make sure the path CARBON_HOME does not contain spaces."
```
ashendes, line 181: This message is not clear as the user will be not aware of CARBON_HOME. Lets use a generic message.

PR #20366, opened 2024-05-13
Bump framework and oauth versions

sadilchamishka:
$subject

### DIFF
--- a/modules/distribution/pom.xml
+++ b/modules/distribution/pom.xml
@@ -585,6 +585,7 @@
                 <artifactId>maven-site-plugin</artifactId>
                 <version>3.3</version>
                 <configuration>
+
                     <reportPlugins>
                         <plugin>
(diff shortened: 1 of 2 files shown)


modules/distribution/pom.xml
```diff
@@ -585,6 +585,7 @@
                 <artifactId>maven-site-plugin</artifactId>
                 <version>3.3</version>
                 <configuration>
+
```
SujanSanjula96, line 588: is this intended?

Avarjana (approved):

modules/distribution/pom.xml
sadilchamishka: Fixed by https://github.com/wso2/product-is/pull/20366/commits/c810f9eac0b461388e620fa98005f7ab9e66d9c6
Should be a mistake

SujanSanjula96 (approved):

PR #20430, opened 2024-05-21
Bump carbon registry version to 4.8.34

shanggeeth:
$Subject

### Purpose
- Bump the packed Jetty version to 10.0.20 via upgrading to solr 9.5.0.wso2v6


ZiyamSanthosh (approved):

darshanasbg:
@shanggeeth: I came across https://github.com/wso2/product-is/pull/20254 and as per the offline chat with @UdeshAthukorala there were some complications updating registry dependency.. 

Not sure whether those are resolved now, but in case PR builder fails with this PR also, please sync with @UdeshAthukorala as he may have some info on the previous analysis.

ZiyamSanthosh (approved):

PR #20631, opened 2024-07-02
Update pom.xml

VivekVinushanth:


VivekVinushanth:
The only change is from https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2509 has run the PR builder.

kayathiri4 (approved):

PR #21045, opened 2024-09-03
Add comprehensive integration tests for Authorization Code Grant for JWT Access Tokens

malithie:
Resolves https://github.com/wso2/product-is/issues/21046

This PR incorporates integration tests for Authorization Code grant.
- Rather than relying on the playground sample app this test case directly invokes the authorization server for login
- Validate audience, expiry, user claims, and standard claims of the access token
- Trigger the refresh token flow and validate audience, expiry, user claims, and standard claims of the access token obtained with refresh

### DIFF
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/oauth2/OAuth2RefreshGrantJWTTokenTestCase.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/oauth2/OAuth2RefreshGrantJWTTokenTestCase.java
@@ -1,0 +1,222 @@
+/*
+ * Copyright (c) 2024, WSO2 LLC. (http://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/oauth2/OAuth2RefreshGrantJWTTokenTestCase.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+import static org.wso2.identity.integration.test.utils.OAuth2Constant.ACCESS_TOKEN_ENDPOINT;
+import static org.wso2.identity.integration.test.utils.OAuth2Constant.AUTHORIZATION_HEADER;
+
+public class OAuth2RefreshGrantJWTTokenTestCase extends OAuth2ServiceAbstractIntegrationTest {
```
Thumimku, line 57: Is this class added to testng file?

malithie: No it's kept as this so we can invoke from different grants like password etc. As an additional improvement I'm looking at invoking this programmatically as a test class. That's why it's added with test annotations.

Will do that improvement separately

Thumimku (approved):

PR #21050, opened 2024-09-04
Bump Framework and api server versions

ashanthamara:


ashanthamara:
Closing this since the same dependencies were bumped by https://github.com/wso2/product-is/pull/21051 and there is no need to add framework dependencies to the root pom unless they are directly used in the repo.

PR #21075, opened 2024-09-08
Change_Port_Number

Farhan-labib:
When integrating WSO2 with another project, you might encounter a common issue if both the WSO2 server and the other project are configured to use the same port. To resolve this, you need to change the default port used by WSO2. Here’s a clear guide on how to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,13 @@ By default mode, it would ask you to enter the master password
 
 > ciphertool -Dchange  (in UNIX)
 
+4. Change the Port by modifying the conf/axis2.xml file
+
+Locate the `<offset>` tag in the conf/axis2.xml file.
+Change the value from 0 to 2 (or the desired offset).
+> For example: `<offset>2</offset>`
+> With this change, if the default port is 9443, it will now run on port 9443 + 2 = 9445.
+
 For more details see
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -195,6 +195,13 @@ By default mode, it would ask you to enter the master password
 
 > ciphertool -Dchange  (in UNIX)
 
+4. Change the Port by modifying the conf/axis2.xml file
```
madurangasiriwardena, line 198: Do you mean the carbon.xml?

Even so, manually changed the configuration will be overridden with the configurations in the deployment.toml.

Correct way to change the port numbers are mentioned in the product documentation https://is.docs.wso2.com/en/latest/references/default-ports/#change-the-offset-for-default-ports

madurangasiriwardena:
@Farhan-labib,

Thank you for your contributions. As mentioned in the previous comment, correct way to change the port number is mentioned below, which is extracted from the product documentation.

There are two ways to set an offset to a port:

- Pass the port offset to the server during startup. The following command starts the server with the default port incremented by 3.
  `./wso2server.sh -DportOffset=3`
- Set the offset value in the <IS_HOME>/repository/conf/deployment.toml file as follows:
  ```
  [server]
  offset = <PORT_OFFSET_VALUE>
  ```

Closing the PR since this already documented in the product documentation.

PR #21152, opened 2024-09-24
Bump kernel to 4.10.22

hwupathum:
$subject


UdeshAthukorala (approved):

darshanasbg:
Have we run the integration tests with this change? product-is master tests failing.. Can we cross check and confirm this change doesn't caused for that?

Issue #5661, opened 2019-06-12
[Sample] Add a sample to support OIDC back-channel logout


hasinidilanka:
Add OIDC back-channel logout support to a sample application. 

Related to : https://wso2.org/jira/browse/IDENTITY-6596

isharak:
This issue is being closed due to extended inactivity. Please feel free to reopen it if further attention is needed. Thank you for helping us keep the issue list relevant and focused!

PR #21294, opened 2024-10-08
Fix SCIM custom claim dialect related test issue

PasinduYeshan:
Fix:
The SCIM2CustomSchemaUserTestCase.createClaims() test is failing with a NoSuchMethodError:
```
[ERROR] org.wso2.identity.integration.test.scim2.rest.api.customSchema.SCIM2CustomSchemaUserTestCase.createClaims(org.wso2.identity.integration.test.scim2.rest.api.customSchema.SCIM2CustomSchemaUserTestCase)
[ERROR] Run 1: SCIM2CustomSchemaUserTestCase.createClaims:153 » NoSuchMethod org.wso2.carbon....
[ERROR] Run 2: SCIM2CustomSchemaUserTestCase.createClaims:153 » NoSuchMethod org.wso2.carbon....
```

### Root Cause
This error is occurring due to changes introduced in PR #5661 (https://github.com/wso2/carbon-identity-framework/pull/5661). That PR added claims to the "urn:scim:wso2:schema" dialect by default. However, the SCIM2CustomSchemaUserTestCase assumes this custom dialect URI is not set and attempts to create the claim dialect within the test.

### Solution
Updated test to, check if custom schema dialect exists and create dialect only if it doesn't exist.

is-tests-scim2 tests passed locally with this change.
image

### DIFF
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/scim2/rest/api/customSchema/SCIM2CustomSchemaMeTestCase.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/scim2/rest/api/customSchema/SCIM2CustomSchemaMeTestCase.java
@@ -21,6 +21,7 @@ package org.wso2.identity.integration.test.scim2.rest.api.customSchema;
 import io.restassured.RestAssured;
 import io.restassured.response.ExtractableResponse;
 import io.restassured.response.Response;
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.http.HttpHeaders;
@@ -46,6 +47,7 @@ import org.wso2.identity.integration.common.clients.claim.metadata.mgt.ClaimMeta
 import org.wso2.identity.integration.test.scim2.rest.api.SCIM2BaseTest;
 
 import java.rmi.RemoteException;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
 
 import static org.hamcrest.CoreMatchers.notNullValue;
@@ -125,7 +127,10 @@ public class SCIM2CustomSchemaMeTestCase extends SCIM2BaseTest {
     public void testFinish() {
 
         try {
-            claimMetadataManagementServiceClient.removeClaimDialect(CUSTOM_SCHEMA_URI);
+            claimMetadataManagementServiceClient.removeExternalClaim(CUSTOM_SCHEMA_URI, COUNTRY_CLAIM_ATTRIBUTE_URI);
+            claimMetadataManagementServiceClient.removeExternalClaim(CUSTOM_SCHEMA_URI,
+                    MANAGER_EMAIL_CLAIM_ATTRIBUTE_URI);
+            claimMetadataManagementServiceClient.removeExternalClaim(CUSTOM_SCHEMA_URI, MANAGER_CLAIM_ATTRIBUTE_URI);
             claimMetadataManagementServiceClient.removeLocalClaim(MANAGER_LOCAL_CLAIM_URI);
         } catch (RemoteException | ClaimMetadataManagementServiceClaimMetadataException e) {
            log.error(e);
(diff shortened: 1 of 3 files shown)


ashensw:
@PasinduYeshan Any idea why it was giving the error as `NoSuchMethod`, which seems a bit confusing?

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/scim2/rest/api/customSchema/SCIM2CustomSchemaUserTestCase.java
```diff
@@ -145,17 +146,25 @@ private void createClaims() throws Exception {
...
+        }
+
+        //Set claims.
         setSimpleAttribute();
         setComplexAttribute();
-        assertEquals(claimMetadataManagementServiceClient.getClaimDialects().length, claimDialect + 1);
+
+        ClaimDialectDTO[] updatedClaimDialects = claimMetadataManagementServiceClient.getClaimDialects();
+        Assert.assertTrue(Arrays.stream(updatedClaimDialects)
+                .anyMatch(dialect -> StringUtils.equals(dialect.getClaimDialectURI(), CUSTOM_SCHEMA_URI)));
```
ashensw, line 167: Since the method is about creating claims, shall we add an assertion to check the claims addition as well, apart from validating the dialect?

PasinduYeshan:
> Any idea why it was giving the error as `NoSuchMethod`, which seems a bit confusing?

After analyzing the bytecode, it looks like the error is actually coming from the error handling code itself, specifically this line:
`Object messageObject = this.fromOM(faultElt, messageClass, (Map)null);`

Problem is likely related to deserializing the SOAP fault response, rather than the main method call.

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/scim2/rest/api/customSchema/SCIM2CustomSchemaMeTestCase.java
```diff
@@ -147,17 +152,33 @@ private void createClaims() throws Exception {
...
+        ClaimDialectDTO[] claimDialects = claimMetadataManagementServiceClient.getClaimDialects();
 
-        // Set custom schema dialect.
-        ClaimDialectDTO claimDialectDTO = new ClaimDialectDTO();
-        claimDialectDTO.setClaimDialectURI(CUSTOM_SCHEMA_URI);
-        claimMetadataManagementServiceClient.addClaimDialect(claimDialectDTO);
+        boolean isCustomSchemaDialectExist = Arrays.stream(claimDialects)
+                .anyMatch(claimDialect -> StringUtils.equals(claimDialect.getClaimDialectURI(), CUSTOM_SCHEMA_URI));
+
+        if (!isCustomSchemaDialectExist) {
```
madurangasiriwardena, line 160: If the claim dialect is existing, the purpose of this testcase will not be fulfilled... What is the possibility of this being already available? I believe, we need to avoid this check.

PasinduYeshan: With this PR (https://github.com/wso2/carbon-identity-framework/pull/5661), we register the custom claim dialect and add external claims. In the test case it tries to create the same claim dialect again, which leads to the issue.

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/scim2/rest/api/customSchema/SCIM2CustomSchemaUserTestCase.java
PasinduYeshan: Added

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/scim2/rest/api/customSchema/SCIM2CustomSchemaMeTestCase.java
PasinduYeshan: Previously, the custom claim dialect was only created in IS when a custom SCIM attribute was added. With the above PR, some SCIM attributes are now packed by default under the custom SCIM claim dialect. As a result, the test case is trying to recreate an already existing dialect, which is causing the issue. I think now we can verify whether the custom test attributes are correctly created within the existing custom claim dialect.

madurangasiriwardena: Either we check the custom dialect addition or check the already existing dialect. We cannot do both here.

PasinduYeshan: Rather than checking custom dialect addition or existence, focus here is moved to test if following custom external claims are properly created within the custom claim dialect, as this test method not only create the claim dialect but also add following external claims to the custom claim dialect.
 - urn:scim:wso2:schema:country
 - urn:scim:wso2:schema:manager
 - urn:scim:wso2:schema:manager.email
 
```
Assert.assertTrue(Arrays.stream(externalClaimDTOs)
                .anyMatch(claim -> StringUtils.equals(claim.getExternalClaimURI(), COUNTRY_CLAIM_ATTRIBUTE_URI)));
                
Assert.assertTrue(Arrays.stream(externalClaimDTOs)
        .anyMatch(claim -> StringUtils.equals(claim.getExternalClaimURI(), MANAGER_CLAIM_ATTRIBUTE_URI)));
        
Assert.assertTrue(Arrays.stream(externalClaimDTOs)
                .anyMatch(claim -> StringUtils.equals(claim.getExternalClaimURI(), MANAGER_EMAIL_CLAIM_ATTRIBUTE_URI)));
```

PasinduYeshan: "urn:scim:wso2:schema" claim dialect adding part is removed, and test description is updated now properly.

sadilchamishka (approved):

Issue #21158, opened 2024-09-25
[Java 21 runtime] Java 21 runtime support for WSO2 Identity Server


indeewari:
**Describe the issue:**
The latest LTS release of Java; Java 21 was released on September 2023. As Identity Server is a downloadable product which will be used years after its releases, its important to plan for the latest LTS support for a upcoming release proactively.

This initiative aims to ensure compatibility with the Java 21 runtime for the IS product.

1. Remove sun.java dependencies to ensure compatibility with the Java 21 LTS 
2. Resolving Strong Encapsulation Errors
3. Product IS builder github actions for 17 and 21
4. Test failures with Nashorn
5. ..

indeewari:
1. Remove sun.java dependencies to ensure compatibility with the Java 21 LTS runtime.

With the introduction of Java 21 LTS, most of the sun.java classes and methods are no longer supported. These internal APIs were never intended for public use and have been strongly encapsulated since Java 9. Continuing to rely on them in Java 21 poses significant risks, including potential runtime failures and compatibility issues. Therefore, it's essential to refactor the codebase to replace sun.java dependencies with standard, supported Java APIs to ensure long-term stability and maintainability of the product.

Related PRs
- https://github.com/wso2/carbon-multitenancy/pull/271

indeewari:
2. Resolving Strong Encapsulation Errors in JDK Internals

Some non-critical JDK internals, which are strongly encapsulated, are accessed via reflection for audit purposes. There are two approaches to address this:

1. Eliminate the use of strongly encapsulated JDK internals accessed through reflection.
2. Opt for relaxed strong encapsulation.

The choice between these options should be made on a case-by-case basis.

After evaluating backward compatibility, impact areas, and implementation complexity, we have opted for relaxed strong encapsulation for the following modules:

For `java.base`:
- `sun.security.util`
- `java.security`
- `sun.security.rsa`
- `java.security.cert`

For `java.rmi`:
- `sun.rmi.transport`

indeewari:
3. Product IS builder github actions for 17 and 21

Java 17 runtime action for IS builder

- https://github.com/wso2/product-is/pull/21179
- https://github.com/wso2/product-is/pull/21183

Java 21 runtime action for IS builder

indeewari:
4. Test failures with Nashorn

- https://github.com/wso2/product-is/pull/21259

indeewari:
5. Update the Docker images on Java 21 runtime.

[...]

PR #21366, opened 2024-10-15
Adding JDK 21 builder

indeewari:
Introducing a new builder for JDK 21 runtime with integration tests.

Related issue
- https://github.com/wso2/product-is/issues/21158

### DIFF
--- a/.github/workflows/product-is-builder-jdk21.yml
+++ b/.github/workflows/product-is-builder-jdk21.yml
@@ -1,0 +1,59 @@
+name: product-is-builder-jdk21
+
+on:
+  workflow_dispatch:
+  schedule:
+#    At 00:00 on day-of-month 1
+    - cron:  '0 0 1 * *'
+
+
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/product-is-builder-jdk21.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Checkout repository code
+        uses: actions/checkout@v2
+      - name: Set up Adopt JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: "11"
+          distribution: "adopt"
+      - name: Product-IS build with JDK 11 without Integration Tests
+        run: |
+          mvn clean install --batch-mode -Dmaven.test.skip=true | tee mvn-build.log
```
hwupathum, line 31: Why do we run integration tests with both JDK 11 and 21?

indeewari: -Dmaven.test.skip=true will skip the tests. We are building the pack on JDK 11 in this step. Next step is running the only the integration tests on JDK 21

hwupathum (approved):

PR #21583, opened 2024-11-02
Update framework version

piraveena:
$subject


piraveena:
Skipping pr builder since this includes only one change from framework repo, and that change already succeeded with pr builder.

Avarjana (approved):

PR #21695, opened 2024-11-13
Update git actions

madurangasiriwardena:
- Remove the deprecated set-output command. Refer https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- Remove unused scripts


.github/workflows/dependency-updater.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Update Dependencies
         id: builder_step
         run: |
-          echo "::set-output name=REPO_NAME::${{ env.REPOSITORY }}"
+          echo "REPO_NAME=${{ env.REPOSITORY }}" >> $GITHUB_OUTPUT
```
ayeshajay, line 53:
```suggestion
          echo "REPO_NAME=${{ env.REPOSITORY }}" >> "$GITHUB_OUTPUT"
```

.github/scripts/pr-builder.sh
```diff
@@ -23,7 +23,7 @@ PULL_NUMBER=$(echo $PR_LINK | awk -F'/' '{print $7}')
 echo "    USER: $USER"
 echo "    REPO: $REPO"
 echo "    PULL_NUMBER: $PULL_NUMBER"
-echo "::set-output name=REPO_NAME::$REPO"
+echo "REPO_NAME=$REPO" >> $GITHUB_OUTPUT
```
ayeshajay, line 26:
```suggestion
echo "REPO_NAME=$REPO" >> "$GITHUB_OUTPUT"
```

Adding quotes around $GITHUB_OUTPUT ensures that special characters in the file path do not cause unexpected behavior or errors.

ayeshajay (approved):

PR #21748, opened 2024-11-18
Bump the identity portal version.

Malith-19:
$subject


ashensw:
Versions were updated with https://github.com/wso2/product-is/pull/21755

PR #22413, opened 2025-01-24
Move OIDCPasswordGrantTest from test block with restarts

madurangasiriwardena:


janakamarasena (approved):

madurangasiriwardena:
Only the RegistryMountTestCase test case is failing, which is failing already. Hence merging the PR.

PR #23092, opened 2025-02-17
Create label-doc-issues.yml

isharak:
Automatically labeling docs-related issues with the correct issue type

### DIFF
--- a/.github/workflows/label-doc-issues.yml
+++ b/.github/workflows/label-doc-issues.yml
@@ -1,13 +1,3 @@
-# -------------------------------------------------------------------------------------
-#
-# Copyright (c) 2025, WSO2 LLC. (http://www.wso2.com). All Rights Reserved.
-#
-# This software is the property of WSO2 LLC. and its suppliers, if any.
-# Dissemination of any information or reproduction of any material contained
-# herein in any form is strictly forbidden, unless permitted by WSO2 expressly.
-# You may not alter or remove any copyright or other notice from copies of this content.
-#
-# --------------------------------------------------------------------------------------
 
 name: Label New Document Issues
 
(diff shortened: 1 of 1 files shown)


.github/workflows/label-doc-issues.yml
```diff
@@ -0,0 +1,75 @@
...
+            }
+      
+      - name: Read issue content and determine label
+        id: determine_label
+        if: steps.check_label.outputs.has_docs_label == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const issue = context.payload.issue;
+            const body = issue.body || "";
```
ayeshajay, line 45: Instead of using the body, let's use the labels to get the existing labels.
`const labels = issue.labels.map(label => label.name);`

You can use the following code block for the logic.

```
const labels = issue.labels.map(label => label.name);
const validLabels = ["Type/Bug", "Type/Improvement", "Type/NewFeature", "Type/Task"];
let assignedLabel = "Type/Bug"; // Default label

for (const label of labels) {
  if (validLabels.includes(label)) {
    assignedLabel = label;
    break;
  }
}
core.setOutput("label", assignedLabel);
```

isharak: In this stage only "Type/Docs" label available in issue labels, other types will be in the issue body.
Please check this git issue.
https://github.com/wso2/product-is/issues/23096

ayeshajay (dismissed):

.github/workflows/label-doc-issues.yml
```diff
@@ -0,0 +1,65 @@
+
```
ayeshajay, line 1: Remove the empty line at the top.

ayeshajay (approved):

PR #23142, opened 2025-02-18
Bump nimbusds version to 9.37.3

sandushi:
### Purpose
$subject
- Fix tests cases


sandushi:
The PR is still open because this is a major version upgrade and need more testing effort, will merge this PR after the testing effort.

sandushi:
This upgrade will be done as  another effort; hence, we are closing this issue.

PR #23176, opened 2025-02-20
Bump identity-apps component versions

DilshanSenarath:


DilshanSenarath:
The PR builder was triggered in the fork: https://github.com/DilshanSenarath/product-is/actions/runs/13427915104  

There is only one test failure, which is an intermittently failing test case.
image

PR #23970, opened 2025-05-06
Fix intermittent test failures for the secondary user store adding.

Malith-19:
### Purpose
- $subject

### Related issues
- https://github.com/wso2/product-is/issues/24003

### DIFF
--- a/modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/identity/integration/common/utils/UserStoreConfigUtils.java
+++ b/modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/identity/integration/common/utils/UserStoreConfigUtils.java
@@ -40,6 +40,7 @@ public class UserStoreConfigUtils {
     public boolean waitForUserStoreDeployment(UserStoreConfigAdminServiceClient userStoreConfigAdminServiceClient,
                                               String domain) throws Exception {
 
+        Thread.sleep(5000);
         long waitTime = System.currentTimeMillis() + 30000; //wait for 45 seconds
         while (System.currentTimeMillis() < waitTime) {
             UserStoreDTO[] userStoreDTOs = userStoreConfigAdminServiceClient.getActiveDomains();
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/server/user/store/v1/UserStoreSuccessTest.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/server/user/store/v1/UserStoreSuccessTest.java
@@ -1,17 +1,19 @@
 /*
- * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2019-2025, WSO2 LLC. (http://www.wso2.com).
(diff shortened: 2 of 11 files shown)


PasinduYeshan (approved):

amanda-ariyaratne (dismissed):

Malith-19:
Test runner has passed in the fork https://github.com/Malith-19/product-is/actions/runs/14939836135

modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/identity/integration/common/utils/UserStoreConfigUtils.java
```diff
@@ -40,6 +40,7 @@ public class UserStoreConfigUtils {
     public boolean waitForUserStoreDeployment(UserStoreConfigAdminServiceClient userStoreConfigAdminServiceClient,
                                               String domain) throws Exception {
 
+        Thread.sleep(5000);
         long waitTime = System.currentTimeMillis() + 30000; //wait for 45 seconds
```
pavinduLakshan, line 44: just asking to be aware, cannot we do this instead? 

```suggestion
        long waitTime = System.currentTimeMillis() + 35000; //wait for 35 seconds
```

Malith-19: This wait time is the time we will be used to do the configuration check timeout. That means we are waiting maximum of 30 seconds. 

The first sleep is used to wait before the first configuration check API call. Since this is a user store adding it will never configure just after we send the create request. So if we remove the first 5s wait, it will lead to some additional API calls and intension here is to reduce those redundant API calls.

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/server/user/store/v1/UserStoreSuccessTest.java
```diff
@@ -52,9 +53,9 @@
 public class UserStoreSuccessTest extends UserStoreTestBase {
 
     private static String domainId;
-    private UserStoreConfigAdminServiceClient userStoreConfigAdminServiceClient;
-    private UserStoreConfigUtils userStoreConfigUtils;
+    private UserStoreMgtRestClient userStoreMgtRestClient;
     private static final String USER_STORE_TYPE_ID = "VW5pcXVlSURKREJDVXNlclN0b3JlTWFuYWdlcg";
+    private static final String USER_STORE_DOMAIN = "JDBC-3";
```
ashensw, line 58: Can we change the userstore name to a better one?

Malith-19: This name is coming from a config file, https://github.com/wso2/product-is/blob/17559d06d5fd2f0b108b1e8ae1ad80db118fc333/modules/integration/tests-integration/tests-backend/src/test/resources/org/wso2/identity/integration/test/rest/api/server/user/store/v1/user-store-add-secondary-user-store.json#L4

ashensw (approved):

PR #24140, opened 2025-05-27
Bump org-mgt version

Yasasr1:
$subject

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2602,7 +2602,7 @@
         <identity.extension.utils>1.0.21</identity.extension.utils>
         <authenticator.auth.otp.commons.version>1.0.11</authenticator.auth.otp.commons.version>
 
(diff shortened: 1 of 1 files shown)


Yasasr1:
The PR builder is failing because the changes introduced in the following PR have not yet been merged into the product 
- https://github.com/wso2-extensions/identity-event-handler-notification/pull/325

As a result, the PR builder will also fail for that PR, since it depends on the changes in this PR. These two PRs are interdependent, and both need to be merged for the PR builder to pass successfully.

I have tested the changes locally using a pack that includes both sets of changes and confirmed that all failing tests pass. For your reference, I’ve attached the surefire report [1].

We need to proceed by first merging this PR, followed by the above mentioned PR. Please note that until both PRs are merged, the integration tests will continue to fail.

[1] - 
[surefire-report.zip](https://github.com/user-attachments/files/20455909/surefire-report.zip)

Issue #24289, opened 2025-06-11
Remove expired certificates from the default key store


madurangasiriwardena:
### Current Limitation

This log line can be seen at the server startup.

`[2025-06-11 11:43:39,210] []  WARN {org.apache.tomcat.util.net.SSLUtilBase} - The trusted certificate with alias [cn_baltimore_cybertrust_root,ou_cybertrust,o_baltimore,c_ie] and DN [CN=Baltimore CyberTrust Root, OU=CyberTrust, O=Baltimore, C=IE] is not valid due to [NotAfter: Tue May 13 05:29:00 IST 2025]. Certificates signed by this trusted certificate WILL be accepted`

### Suggested Improvement

We need to remove this expired certificate from the default key store and possibly sync with key store from the JDK.

### Version

7.2.0 m3

PR #24293, opened 2025-06-12
Remove expired certificates from the key store

hwupathum:
$subject

Since the current latest Java version (Java 24) has the same expired certificate. We are removing the expired cert from keystore.

Related issue https://github.com/wso2/product-is/issues/24289


hwupathum:
PR builder is failing because we are adding binaries in the PR. I have verified the integration tests locally.

UdeshAthukorala (approved):

PR #24391, opened 2025-06-20
Update README to use dynamic product version in documentation links

KD23243:
## Description
The `README.txt` file bundled in the latest release ZIP contains outdated documentation references, including hardcoded version numbers (e.g., 6.0.0), incorrect server URLs, and an old copyright year.

## Proposed Solution
- Replace hardcoded version numbers in documentation links with ${product.version}.
- Update web console URL to https://localhost:9443/console/.
- Update copyright year.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -16,9 +16,9 @@ seamless modularity to your SOA solution via componentization.
 System Requirements
 ===================
 
-1. Minimum memory - 2 GB
+1. Minimum memory - 4 GB
 
-2. Processor - 2 Core/vCPU 1.1GHz or higher.
+2. Processor - 4 vCPUs (x86_64 Architecture) or higher.
 
 3. Java SE Development Kit 11
 
@@ -27,7 +27,7 @@ System Requirements
 5. To build WSO2 Identity Server from the Source distribution, it is also necessary that you have Maven 3 or later.
 
 For more details see
-    https://is.docs.wso2.com/en/${product.version}/deploy/get-started/install/
+    https://is.docs.wso2.com/en/latest/deploy/get-started/install/
 
 
 Project Resources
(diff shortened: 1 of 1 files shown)


README.txt
```diff
@@ -27,15 +27,15 @@ System Requirements
 5. To build WSO2 Identity Server from the Source distribution, it is also necessary that you have Maven 3 or later.
 
 For more details see
-    https://is.docs.wso2.com/en/6.0.0/deploy/get-started/install/
+    https://is.docs.wso2.com/en/${product.version}/deploy/get-started/install/
```
Malith-19: Instead of a variable can't we use latest here?

KD23243: This will depend on how we decide to approach documentation in the future right? Currently 7.1.0 redirects to 'latest'. However if we decide to move away from using 'latest', we might need to revisit and update these. Either approach will work.

bhagyasakalanka: How will this be replaced? `${product.version}` The link will not work if we keep a placeholder.

KD23243: Updated to latest

README.txt
```diff
@@ -172,7 +172,7 @@ By default mode, it would ask you to enter the master password
 > ciphertool -Dchange  (in UNIX)
 
 For more details see
-https://is.docs.wso2.com/en/6.0.0/deploy/security/encrypt-passwords-with-cipher-tool/
+https://is.docs.wso2.com/en/latest/deploy/security/encrypt-passwords-with-cipher-tool/
```
hwupathum, line 175: Using the latest tag in the readme will be outdated when the next IS version is released. IMO we should keep this as 7.1.0 for now

nilasini: Yes. When we release a new version the document will only be published after 2 to 3 weeks only hence we have to add this as a post release task and do after each releases

PR #24544, opened 2025-07-04
Bump api-server

Sachin-Mamoru:
This pull request includes a minor version update to the `identity.server.api.version` in the `pom.xml` file, changing it from `1.3.141` to `1.3.142`.


indeewari (approved):
Please merge only when the version is bumped at `api-server`

PR #24874, opened 2025-07-28
Add integration test for SCIM2 Roles V3 API

asha15:
$subject


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/scim2/SCIM2RolesV3TestCase.java
```diff
@@ -0,0 +1,330 @@
+/*
+ * Copyright (c) 2025, WSO2 Inc. (http://www.wso2.org).
```
madurangasiriwardena, line 2: License header is not correct.

Issue #24983, opened 2025-08-08
Unable to get configured attributes in Pre Update Password action in Registration Flows


ashanthamara:
### Description

We can configure the required attributes that need to be sent to the extension in Pre update password action.
But however after improving the action to support password validation in registration flows, attribute resolving part is not functioning as the logic was to check the claims from the userstore.
Since the action executes at Pre user add level, retrieving userclaims doesnt work in that context.

### Steps to Reproduce

.

### Please select the area issue is related to

User & Identity Administration

### Version

7.2

### Developer Checklist

- [ ] [Behavioural Change] Does this change introduce a behavioral change to the product?
- [ ]  ↳ Approved by team lead
- [ ]  ↳ Label `impact/behavioral-change` added
- [ ] [Migration Impact] Does this change have a migration impact?
- [ ]  ↳ Migration label added (e.g., `7.2.0-migration`)
- [ ]  ↳ Migration issues created and linked
- [ ] [New Configuration] Does this change introduce a new configuration?
- [ ]  ↳ Label `config` added
- [ ]  ↳ Configuration is properly documented

PR #24984, opened 2025-08-08
Fix integration tests related password action execution in user registration

ashanthamara:
$subject

### Related Issue
- https://github.com/wso2/product-is/issues/24983


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/serviceextensions/common/ActionsBaseTestCase.java
```diff
@@ -66,10 +66,34 @@ protected void init(TestUserMode userMode) throws Exception {
     public String createAction(String actionType, ActionModel actionModel) throws IOException {
 
         String actionId = actionsRestClient.createActionType(actionModel, actionType);
-        actionsRestClient.activateAction(actionType, actionId);
+        activateAction(actionType, actionId);
```
indeewari, line 69: NO need to validate if actionId is valid  != -1 or !=0 ?

indeewari (dismissed):

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/serviceextensions/common/ActionsBaseTestCase.java
ashanthamara: This method is used as a utility method to create and activate action. So no need to validate the Id.

PR #25759, opened 2025-09-24
Remove Bcrypt from dropins and add it to plugins.

oshinika:


pavinduLakshan:
Let's keep the PR template and update it accordingly.

Issue #24222, opened 2025-06-03
lang attribute doesn't change in some pages when language is switched


pabasara-mahindapala:
### Description

The `lang` attribute in the HTML tag is not changing when the language is switched in pages like `verify.do` and `completepasswordreset.do`. This seems to be due to the `lang` attribute being set in the <IS_HOME>/repository/deployment/server/webapps/authenticationendpoint/includes/language-switcher.jsp file. In the files without the language switcher, the `lang` attribute is not correctly set to the selected language.

### Steps to Reproduce

Steps to reproduce:

1. Follow the reset password flow after changing the language.
2. lang attribute in the `verify.do` page will be set to en-US even when the selected language is different.

![Image](https://github.com/user-attachments/assets/b4b1f544-e73b-45f6-afe9-36270e78e462)

### Version

IS 7.1.0

CodeWithNed:
Hi,
I’d like to take this issue. Could you please assign it to me?

Thanks!

pavinduLakshan:
Hi @CodeWithNed, I have assigned you to the issue. Please ping if you need any support. Thanks!

sagara-gunathunga:
@CodeWithNed Just checking in to see if you’re still working on this issue. If you are, it would be great to share a quick update or a draft PR to show your progress. If you’re no longer working on it, please let us know so we can make the issue available to other participants.

sagara-gunathunga:
Due to a long period of inactivity, we have unassigned this issue. It is now available for other contributors to pick up.

KenulaNimhan:
Hello,

I would like to work on this issue. Can you assign it to me ? @sagara-gunathunga

Thanks!

pavinduLakshan:
Hi @KenulaNimhan, assigned to you. Let's aim to open a pull request within the week. Please don't hesitate to reach out if you have any concerns. Good luck!

KenulaNimhan:
Hi @pavinduLakshan 

I’ve been looking into this issue and noticed that the affected JSP (language-switcher.jsp) lives in the [identity-apps repository ](https://github.com/wso2/identity-apps) under,

identity-apps-core/apps/recovery-portal/src/main/webapp/includes/language-switcher.jsp

I plan to implement the fix there and open a PR against identity-apps.

Please confirm if that’s the correct place for this change.

pavinduLakshan:
> Hi [@pavinduLakshan](https://github.com/pavinduLakshan)
> 
[...]

PR #25771, opened 2025-09-24
Fix: dynamically set html lang on all auth pages

Hiruja99k:
**Description**

This PR fixes the issue where the `lang` attribute in the HTML `<html>` tag did not update correctly on some pages, such as `verify.do` and `completepasswordreset.do`, after switching languages.

Previously, the `lang` attribute was only updated by `language-switcher.jsp`. Pages that did not include this JSP remained stuck at `en-US` even when a different language was selected.

**What was done**
- Modified the distribution module (`modules/distribution/pom.xml`) to ensure all `authenticationendpoint` and `accountrecoveryendpoint` pages dynamically set the `<html lang>` attribute.
- The `lang` is now set from the selected `ui_locales` parameter, or falls back to the request locale if `ui_locales` is not provided.
- Any pre-existing `lang` attribute is removed first to avoid duplicates.

**Impact**
- All pages in the authentication and account recovery flows now correctly reflect the selected language.
- Resolves the issue where pages without `language-switcher.jsp` stayed at `en-US` after switching languages.

**Related Issue**
Resolves: #24222

### DIFF
--- a/modules/distribution/pom.xml
+++ b/modules/distribution/pom.xml
@@ -483,6 +483,22 @@
                                 </unzip>
                                 <delete file="../p2-profile-gen/target/wso2carbon-core-${carbon.kernel.version}/repository/deployment/server/webapps/authenticationendpoint.war" />
 
+                                <!-- Ensure correct HTML lang attribute in authenticationendpoint pages -->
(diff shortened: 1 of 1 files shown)


pavinduLakshan:
The ideal fix should be done at the JSP level.

PR #25990, opened 2025-10-09
Add webhook management api integration tests

Sachin-Mamoru:
This pull request introduces comprehensive integration tests for the Webhook Management REST API, covering both successful (happy path) and failure (negative path) scenarios. It adds two new test classes—one for validating correct API behavior and another for ensuring robust error handling. Additionally, a shared test base class is provided to centralize common logic and constants, and the test suite configuration is updated to include these new tests.

**New Webhook Management Integration Tests**

* Added `WebhookManagementSuccessTest` to validate all major happy path operations for the webhook API, including create, list, get, update, activate, deactivate, and delete actions. These tests assert correct status codes, response bodies, and business logic (e.g., secrets are not echoed, channels are handled correctly).
* Added `WebhookManagementFailureTest` to verify negative scenarios such as invalid endpoint URLs, empty event profile fields, and operations on non-existent webhook IDs. These tests assert appropriate error codes, messages, and descriptions for each failure case.

**Test Framework Improvements**

* Introduced `WebhookManagementTestBase`, a shared base class for webhook management tests. It provides common constants, test setup/teardown, utility methods for JSON conversion, and a data provider for user modes, ensuring consistency and reducing duplication.

**Test Suite Configuration**

* Updated `testng.xml` to include both the success and failure webhook management test classes, ensuring they are executed as part of the integration test suite.

### DIFF
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/server/webhook/management/v1/WebhookManagementFailureTest.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/identity/integration/test/rest/api/server/webhook/management/v1/WebhookManagementFailureTest.java
@@ -1,0 +1,199 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (http://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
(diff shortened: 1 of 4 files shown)


Sachin-Mamoru:
fixed via https://github.com/wso2/product-is/pull/25998

Issue #23292, opened 2025-02-27
Sub orgs with special chars in the name become read only


rksk:
### Description

It is possible to create/rename a sub org and use special characters in the same. Once a sub org has such a name, the org edit page becomes read only.

It occurs because the following API call does not return any data when the sub org name has special chars.
```
GET /api/users/v1/me/organizations?authorizedAppName=Console&filter=name+eq+5643$&limit=10&recursive=false
```

### Steps to Reproduce

- Create a sub org with org name `test`
- Goto edit page and rename it with `test$`
- Now if you go back to sub org edit page, it becomes read-only

<img width="399" alt="Image" src="https://github.com/user-attachments/assets/cccc5f5f-891c-4833-88a3-d2503b0cce1e" />

### Version

wso2is-7.1.0-rc2

Jalina2007:
Hi @rksk, I'd like to work on this issue. Can you assign this to me?

pavinduLakshan:
Hi @Jalina2007 , let's attend to this once your current assigned issues are completed.

mustafaazad03:
Hey @pavinduLakshan, can I work on this?

pavinduLakshan:
Hi @mustafaazad03, sure. assigned the issue to you. Please don't hesitate to let us know if you need to clarify anything. Good luck!

JayanaGunaweera01:
Hi,

Can I work on this?

pavinduLakshan:
Hi @JayanaGunaweera01, assigned you the issue, and I tested the fix you sent. While it works as expected for a org name with a single special character (eg: `org$`, `^myorg`), that was not working for an org name with `\` character. Shall we improve the fix to handle that case as well?

JayanaGunaweera01:
> Hi [@JayanaGunaweera01](https://github.com/JayanaGunaweera01), assigned you the issue, and I tested the fix you sent. While it works as expected for a org name with a single special character (eg: `org$`, `^myorg`), that was not working for an org name with `\` character. Shall we improve the fix to handle that case as well?

Hi @pavinduLakshan,

Thank you for pointing that out. I fixed it with [66bdf92](https://github.com/wso2/identity-apps/pull/9362/commits/66bdf922ca60dde4a11f782cd4e75d146a1c0af6)

https://github.com/user-attachments/assets/ca1e60ec-297b-4ed6-965c-639b8402d7f4

sagara-gunathunga:
[...]

PR #26034, opened 2025-10-11
fix: Bug which causes read-only behaviour for a special character

mustafaazad03:
[#23292](https://github.com/wso2/product-is/issues/23292) Resolved the issue for the read only behaviour due to the spacial character


mustafaazad03:
What happend @pavinduLakshan ? Is any issue?

pavinduLakshan:
> What happend @pavinduLakshan ? Is any issue?

Hi @mustafaazad03, could you please share how did you test this fix?

PR #26763, opened 2026-02-18
Update the PR builder to use correct component branches

madurangasiriwardena:
## Summary by CodeRabbit

* **Chores**
  * Updated build/deploy scripts to reference newer branch versions for internal components (kernel, deployment, analytics-common).
  * Simplified analytics-common flow by removing an intermediate extension block and consolidating checkout to the updated branch.
  * Updated console prompts to reflect the new branch/version targets.

### DIFF
--- a/.github/scripts/pr-builder.sh
+++ b/.github/scripts/pr-builder.sh
@@ -169,24 +169,19 @@ else
   cd $REPO
   if [ "$REPO" = "carbon-kernel" ]; then
     echo ""
-    echo "Checking out for 4.10.x branch..."
+    echo "Checking out for 4.12.x branch..."
     echo "=========================================================="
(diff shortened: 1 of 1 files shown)


SujanSanjula96 (dismissed):

.github/scripts/pr-builder.sh
```diff
@@ -169,24 +169,19 @@ else
   cd $REPO
   if [ "$REPO" = "carbon-kernel" ]; then
     echo ""
-    echo "Checking out for 4.10.x branch..."
+    echo "Checking out for 4.12.x branch..."
     echo "=========================================================="
     git checkout 4.10.x
```
Malith-19: we need to change here as well

Issue #27345, opened 2026-03-31
MySQL query error in NotificationTypeDAO


Thisara-Welmilla:
### Description
Failing test grid with MySQL with following error

```
INFO  [org.wso2.carbon.automation.engine.testlisteners.TestManagerListener] - =================== Running the test method org.wso2.identity.integration.test.rest.api.server.email.template.v1.EmailTemplatesNegativeTest.testGetEmailTemplateTypeWithIncorrectTemplateTypeId ===================
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - [2026-03-24 16:13:41,887] [ccc844b9-49e9-4c14-8a64-a3bf6e72dc40] ERROR {org.wso2.carbon.identity.rest.api.server.email.template.v1.core.ServerEmailTemplatesService} - errorCode: ETM-55003 | message: Server encountered an error while retrieving the email template type identified by the given template-type-id.org.wso2.carbon.email.mgt.exceptions.I18nEmailMgtServerException: Error when retrieving 'AccountEnable
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - �I' template type from carbon.super tenant registry.
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - 	at org.wso2.carbon.email.mgt.EmailTemplateManagerImpl.getEmailTemplateType(EmailTemplateManagerImpl.java:254)
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - 	at org.wso2.carbon.email.mgt.EmailTemplateManagerImpl.getEmailTemplateType(EmailTemplateManagerImpl.java:214)
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - 	at org.wso2.carbon.identity.rest.api.server.email.template.v1.core.ServerEmailTemplatesService.getEmailTemplateType(ServerEmailTemplatesService.java:133)
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - 	at org.wso2.carbon.identity.rest.api.server.email.template.v1.impl.EmailApiServiceImpl.getEmailTemplateType(EmailApiServiceImpl.java:112)
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - 	at org.wso2.carbon.identity.rest.api.server.email.template.v1.EmailApi.getEmailTemplateType(EmailApi.java:217)
[...]

PR #27356, opened 2026-03-31
Fix MySQL charset, MSSQL env var, and Ubuntu version in DB test runner

KD23243:
## Summary

- Remove `CHARACTER SET latin1` from MySQL database creation — MySQL 8.0 default `utf8mb4` is now used, matching the app connection charset and surfacing collation mismatches (e.g. wso2/product-is#27345)
- Replace deprecated `SA_PASSWORD` with `MSSQL_SA_PASSWORD` for SQL Server 2022
- Update mssql-tools18 apt repo from `ubuntu/22.04` to `ubuntu/24.04` to match the current `ubuntu-latest` runner, with `-fsSL` added to the curl call to catch HTTP errors

## Test plan

- [ ] Trigger workflow with `database_type=mysql` and verify collation errors surface correctly
- [ ] Trigger workflow with `database_type=mssql` and verify MSSQL starts correctly with the updated env var

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced database initialization reliability with improved startup verification, timeout-based readiness checks, and comprehensive error diagnostics including system logs for easier troubleshooting.
  * Resolved database compatibility concerns by updating character encoding configurations in database creation processes.

* **Chores**
  * Updated CI/CD configuration and package repository references to improve system compatibility across environments.

### DIFF
--- a/.github/workflows/integration-test-runner-mssql.yml
+++ b/.github/workflows/integration-test-runner-mssql.yml
@@ -143,7 +143,7 @@ jobs:
         if: ${{ github.event.inputs.database_type == 'mssql' }}
         run: |
           docker run -d --name mssql \
-            -e SA_PASSWORD="$DB_PASSWORD" \
+            -e MSSQL_SA_PASSWORD="$DB_PASSWORD" \
             -e ACCEPT_EULA=Y \
             -e MSSQL_PID=Developer \
             -p 1433:1433 \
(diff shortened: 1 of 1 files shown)


UdeshAthukorala:
Hi @KD23243 ,

Changing the MySQL integration test runner charset to `utf8mb4` does not align with the actual product behaviour. Within WSO2 Identity Server, **the MySQL database creation scripts explicitly define `latin1` as the charset for the notification template type tables**. Unicode support (`utf8mb4`) was intentionally introduced only for user attribute related tables - email template type name columns are `latin1` by design, since:

- Email template types are system-defined and unicode names are not a supported or encouraged use case.
- Template content (which does support unicode) is stored as BLOB values and is unaffected by this charset.

Running integration tests against a `utf8mb4` database while the product creates tables with `latin1` means the tests are running against a configuration that **does not reflect real-world deployments**, which could mask similar issues in the future.


The root cause was the test itself using a unicode-containing incorrect template type ID. I have addressed this properly by updating the test to use a pure ASCII-only incorrect template type ID, which avoids the collation mismatch entirely and produces the expected `404 Not Found` response. You can find the fix here: https://github.com/wso2/product-is/pull/27374


So could you please **revert the charset change** in the MySQL integration test runner introduced in this PR? With the test-level fix in place, this workaround is no longer needed and keeping it would cause the integration tests to run against a database configuration inconsistent with the actual product.

PR #28298, opened 2026-08-12
Register OpenID4VP presentation definition and VC verification REST APIs in web.xml

Zeta201:
## Summary

Note the $subject


modules/api-resources/api-resources-full/src/main/webapp/WEB-INF/web.xml
```diff
@@ -83,6 +83,8 @@
                 org.wso2.carbon.identity.rest.api.server.workflow.v1.WorkflowAssociationsApi,
                 org.wso2.carbon.identity.rest.api.server.workflow.v1.WorkflowInstancesApi,
                 org.wso2.carbon.identity.api.server.vc.template.management.v1.VcTemplatesApi,
+                org.wso2.carbon.identity.api.server.vp.template.management.v1.PresentationDefinitionsApi,
```
hwupathum, line 86: Need to bump the version

